PR #23, opened 2019-08-23
Convert requests mapping into array

tmpolaczyk:
Convert the `requests` mapping into an array to allow iteration from client code.
For some reason tests fail.

Note: this PR includes code from PRs #19 and #22.


contracts/WitnetBridgeInterface.sol
```diff
@@ -94,12 +94,9 @@ contract WitnetBridgeInterface is VRF {
...
   returns(uint256 _id) {
-    _id = uint256(sha256(_dr));
-    if(requests[_id].dr.length != 0) {
-      requests[_id].tallyReward += _tallyReward;
-      requests[_id].inclusionReward += msg.value - _tallyReward;
-      return _id;
-    }
+    _id = requests.length;
+    DataRequest memory dr;
+    requests.push(dr);
```
aesedepece: Is this kind of initialization needed at all? I suspect this could just work:
```solidity
_id = requests.length;
requests[_id].dr = _dr;
requests[_id].inclusionReward = msg.value - _tallyReward;
```

contracts/WitnetBridgeInterface.sol
```diff
@@ -186,8 +183,9 @@ contract WitnetBridgeInterface is VRF {
     drNotIncluded(_id)
  {
     uint256 drRoot = blockRelay.readDrMerkleRoot(_blockHash);
-    uint256 drHash = uint256(sha256(abi.encodePacked(_id, _poi[0])));
-    if (verifyPoi(_poi, drRoot, _index, _id)) {
+    uint256 drOutputHash = uint256(sha256(requests[_id].dr));
+    uint256 drHash = uint256(sha256(abi.encodePacked(drOutputHash, _poi[0])));
```
aesedepece: :+1:

contracts/WitnetBridgeInterface.sol
```diff
@@ -245,6 +245,12 @@ contract WitnetBridgeInterface is VRF {
     return requests[_id].drHash;
   }
 
+  /// @dev Number of data requests in the WBI.
+  /// @return Returns the number of data requests in the WBI.
+  function numRequests() public view returns(uint256){
```
aesedepece:
```suggestion
  function requestsCount() public view returns(uint256){
```

contracts/WitnetBridgeInterface.sol
tmpolaczyk: I don't think so, accessing elements with index >= array.length results in an "Invalid Opcode" exception.

PR #51, opened 2020-01-28
WIP: implement changes to separate block relay

girazoki:
This PR performs necessary changes to separate the block relay into a separate repository, and making it upgreadble. In short, the WBI should be linked now to a BlockRelayProxy contract, whose address should be invariant. If we would like to change the block relay controller behind the proxy, the owner of the proxy should set the new address to which the proxy should point to. Summary of the changes:

- Remove all BlockRelay contracts.
- Remove block relay specific related tests.
- Add Block Relay Mock contract (centralized), necessary for the tests related to other contracts.
- Import witnet-ethereum-block-relay repository as a depencence.
- Change the migrations to deploy block relay proxy instead of block relay.
- Change all the tests to link the WBI to the block relay proxy contract.
- Add getters for ABS count and membership in the WBI


contracts/ActiveBridgeSetLib.sol
```diff
@@ -1,10 +1,11 @@
 pragma solidity ^0.5.0;
 
-
 /**
  * @title Active Bridge Set (ABS) library
  * @notice This library counts the number of bridges that were active recently
  */
+
+
```
aesedepece: ---

PR #54, opened 2020-02-04
chore: add flattener and update migrations

girazoki:
Adds the flattener file and commands.

Modifies the migrations not to deploy the block relay in networks for which it is already deployed and we have an address.


migrations/addresses.json
```diff
@@ -0,0 +1,8 @@
+{
+  "rinkeby" : {
+    "BlockRelayProxy" : "0xCD835FB8B7359550Cd378f6b048374cE6Da423A8"
+  },
+  "goerli" : {
+  "BlockRelayProxy" : "0xab55369585B03398fDf359D1bAA0D48995C1832A"
```
aesedepece: fmt

PR #62, opened 2020-03-17
fix: rename block-relay in package.json

### LINKED ISSUE #61: Block relay package is being renamed redundantly
There is no reason for renaming the `witnet-ethereum-block-relay` package in `package.json`:
https://github.com/witnet/witnet-ethereum-bridge/blob/dca6d9614004671b6542332a94bf29d44de5d8f1/package.json#L35

This is causing a lot of problems to some users.

To avoid any problems, it should be imported with the same name.

clbartoli:
Rename in package.json and in contracts that import contracts from it:
block-relay ==> witnet-ethereum-block-relay

Closes #61


aesedepece:
While reviewing this, I noticed #61

Issue #68, opened 2020-03-23
Fix errors from code review


girazoki:
# Code Review Result

Tool used: [SmartCheck](https://tool.smartdec.net/)
Commit: 24a7bd3338b9657bc2ee53daaa7e190b059e3bef

## Errors found

Summary:

```
·----------------------------------------------·--------------·
|                Error Type                    |  Occurences  |
·----------------------------------------------·--------------·
| Unsafe array's length manipulation           |           2  |
·······························································
| Multiplication after division                |           1  |
·······························································
| Extra gas consumption                        |           3  |
·······························································
| Costly loop                                  |           6  |
·······························································
| Locked money                                 |           5  |
·······························································
| Compiler version not fixed                   |          25  |
·······························································
| Revert inside the if-operator                |           3  |
·······························································
| Use of SafeMath                              |           4  |
·······························································
| Pure-functions should not read/change state  |           7  |
·······························································
| Non-strict comparison with zero              |           3  |
·······························································
| Prefer external to public visibility level   |          86  |
·······························································
| Use of assembly                              |           7  |
·······························································
| Implicit visibility level                    |          32  |
·----------------------------------------------·--------------·
| TOTAL                                        |         184  |
·----------------------------------------------·--------------·
```

### Unsafe array's length manipulation:

1. File: contracts/CBOR.sol
   Lines: 31-31
   Severity: 1
   Pattern id: 872bdd

2. File: contracts/CBOR.sol
   Lines: 129-129
   Severity: 1
   Pattern id: 872bdd

### Multiplication after division

1. File: contracts/WitnetBridgeInterface.sol
   Lines: 391-391
   Severity: 2
   Pattern id: 09hhh1

### Extra gas consumption

1. File: contracts/WitnetBridgeInterface.sol
   Lines: 167-172
   Severity: 1
   Pattern id: v843m7
   
2. File: contracts/ActiveBridgeSetLib.sol
   Lines: 152-154
   Severity: 1
   Pattern id: v843m7

3. File: contracts/ActiveBridgeSetLib.sol
   Lines: 66-170
   Severity: 1
   Pattern id: v843m7

### Costly loop

1. File: contracts/mocks/MockBlockRelay.sol
   Lines: 199-206
   Severity: 1
   Pattern id: 4b7do5

2. File: contracts/CBOR.sol
   Lines: 36-36
   Severity: 1
   Pattern id: 17f23a

3. File: contracts/CBOR.sol
[...]

PR #69, opened 2020-03-26
[WIP] chore: update versions and fix code review errors

mariocao:
Fix errors from code review #68 : 

```
·----------------------------------------------·--------------·-------·
|                Error Type                    |  Occurences  | FIXED |
······································································|
| Unsafe array's length manipulation           |           2  |     0 |
······································································|
| Multiplication after division                |           1  |     0 |
······································································|
| Extra gas consumption                        |           3  |     3 |
······································································|
| Costly loop                                  |           6  |     0 |
······································································|
| Locked money                                 |           5  |     0 |
······································································|
| Compiler version not fixed                   |          25  |    16 |
······································································|
| Revert inside the if-operator                |           3  |     3 |
·······································································
| Use of SafeMath                              |           4  |     4 |
······································································|
| Pure-functions should not read/change state  |           7  |     0 |
······································································|
| Non-strict comparison with zero              |           3  |     3 |
······································································|
| Prefer external to public visibility level   |          86  |    86 |
······································································|
| Use of assembly                              |           7  |     0 |
······································································|
| Implicit visibility level                    |          32  |    32 |
·----------------------------------------------·--------------·-------·
| TOTAL                                        |         184  |    99 |
·----------------------------------------------·--------------·-------·
```

### Unsafe array's length manipulation: NOT FIXED

These are most probably false negatives caused by mutation of the `length` attribute of a `CBOR.Value memory` data structure, which is wrongly identified by Smartcheck as a length mutation of a dynamic array. This assumption can be validated by renaming `length` as `len` and checking if the same error shows up.

### Multiplication after division: NOT FIXED

Potential precision errors due to a division prior to multiplication. In this case, division is done before in order to avoid overflows. This equation re-arrange was the one leading to less precision/overflow errors.

### Extra gas consumption: FIXED

Potential extra gas consumption due to state variable (e.g. `.length`) of non-memory array is used in the condition of for or while loop.

### Costly loop: NOT FIXED

Comments for errors 1-2:
 - the array length refers to DRs to be claimed
 - the transaction creator is the one interested in using an array of IDs not-long enough to be under the block gas limit

All potentially constly loops in `CBOR.sol` (Errors 4-6) are inherited from the specification in RFC 7049 and are therefore unavoidable. Because of the way in which CBOR is designed to operate, the number of iterations is determined by the value of the first bytes in the `_cborValue.buffer` input.
This could theoretically cause the loop to iterate enough times for the transaction to hit the gas limit. May a third-party contract's business logic depend on this piece of could, it could become unreachable forever. However, there are only two circumstances in which we may find a `_cborValue.buffer` which specified a significantly high length:
1. Ill-intended nonconforming CBOR values that may be specially crafted to force this situation by specifying a declared length that exceeds the actual length of the input bytes. In this case, the algorithm will most surely panic because of shortage of bytes before hitting any gas limit.
2. CBOR values that are actually huge in size. Generally speaking, these loops are indeed not guarded against that possibility. However, specifically in our case, gas limits would be most surely reached by the preceding transaction supplying the input bytes and storing them into `WitnetRequestsBoard`.

### Locked money: NOT FIXED

The SmartDec tool complains about a smart contract having payable methods but without .send, .value, or .transfer methods inside it. In our case, it complains about the Test Helper smart contracts, but these inherit the UsingWitnet.sol contract which has `.value` methods.

### Compiler version not fixed

All have been fixed. Libraries have been instructed to allow any versions from 0.5.3 to less than 0.7.

### Revert inside the if-operator: FIXED

Reverts inside if clauses are not recommended, instead require should be used.

### Use of SafeMath: WILL FIX

According to SmartDec tool, it is a good practice to add overflow checks (SafeMath) only in cases overflow may occur.

### Pure-functions should not read/change state: NOT FIXED

These errors arise when pure functions are potentially changing the state of the contract. Since the tool does not understand assembly language, it does not have a method to derive whether the code is modifying the state. In our cases none of our assembly code is modifying the state of the contracts, but rather input arguments or local variables.

### Non-strict comparison with zero: FIXED

These errors arise when doing some useless comparison, like checking whether an `uint` is bigger or equal than 0.

### Prefer external to public visibility level: FIXED

All functions which have as arguments native solidity types have been renamed external.

### Use of assembly: NOT FIXED

The tool, whenever assembly code is detected, triggers an alarm. These will not be fixed as the assembly code is necessary for the `Buffer` library and a signature verification.

### Implicit visibility level: FIXED

This implies some of the state variables do not have visibility assigned, which by default assigns the public visibility.

### DIFF
--- a/contracts/Request.sol
+++ b/contracts/Request.sol
@@ -2,7 +2,7 @@ pragma solidity >=0.5.3 <0.7.0;
 
 
 /**
- * @title The serialized form of a Witnet data request.
+ * @title The serialized form of a Witnet data request
  */
 contract Request {
   bytes public bytecode;
@@ -14,7 +14,7 @@ contract Request {
   * the WRB should not be considered trustless before a valid Proof-of-Inclusion has been posted for the requests.
   * The hash of the request is computed in the constructor to guarantee consistency. Otherwise there could be a
   * mismatch and a data request could be resolved with the result of another.
-  * @param _bytecode Witnet request in bytes
+  * @param _bytecode Witnet request in bytes.
   */
   constructor(bytes memory _bytecode) public {
     bytecode = _bytecode;
--- a/contracts/Request.sol
+++ b/contracts/Request.sol
@@ -8,12 +8,14 @@ contract Request {
   bytes public bytecode;
   uint256 public id;
 
-  // A `Request` is constructed around a `bytes memory` value containing a well-formed Witnet data request serialized
-  // using Protocol Buffers. However, we cannot verify its validity at this point. This implies that contracts using
-  // the WRB should not be considered trustless before a valid Proof-of-Inclusion has been posted for the requests.
-  //
-  // The hash of the request is computed in the constructor to guarantee consistency. Otherwise there could be a
-  // mismatch and a data request could be resolved with the result of another.
+ /**
+  * @dev A `Request` is constructed around a `bytes memory` value containing a well-formed Witnet data request serialized
+  * using Protocol Buffers. However, we cannot verify its validity at this point. This implies that contracts using
+  * the WRB should not be considered trustless before a valid Proof-of-Inclusion has been posted for the requests.
+  * The hash of the request is computed in the constructor to guarantee consistency. Otherwise there could be a
+  * mismatch and a data request could be resolved with the result of another.
+  * @param _bytecode Witnet request in bytes
+  */
   constructor(bytes memory _bytecode) public {
     bytecode = _bytecode;
     id = uint256(sha256(_bytecode));
--- a/contracts/Request.sol
+++ b/contracts/Request.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.0;
+pragma solidity >=0.5.3 <0.7.0;
 
 
 /**
--- a/contracts/SafeMath.sol
+++ b/contracts/SafeMath.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.6.0;
+pragma solidity 0.6.4;
 
 
 /**
--- a/contracts/SafeMath.sol
+++ b/contracts/SafeMath.sol
@@ -1,0 +1,151 @@
+pragma solidity ^0.6.0;
+
+
+/**
+ * @dev Wrappers over Solidity's arithmetic operations with added overflow
+ * checks.
+ *
+ * Arithmetic operations in Solidity wrap on overflow. This can easily result
+ * in bugs, because programmers usually assume that an overflow raises an
+ * error, which is the standard behavior in high level programming languages.
+ * `SafeMath` restores this intuition by reverting the transaction when an
+ * operation overflows.
+ *
+ * Using this library instead of the unchecked operations eliminates an entire
+ * class of bugs, so it's recommended to use it always.
+ */
+library SafeMath {
+  /**
+   * @dev Returns the addition of two unsigned integers, reverting on
+   * overflow.
+   *
+   * Counterpart to Solidity's `+` operator.
+   *
+   * Requirements:
+   * - Addition cannot overflow.
+   */
+  function add(uint256 a, uint256 b) internal pure returns (uint256) {
+    uint256 c = a + b;
+    require(c >= a, "SafeMath: addition overflow");
+
+    return c;
+  }
+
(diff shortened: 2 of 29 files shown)


contracts/Request.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.0;
+pragma solidity >=0.5.3 <0.7.0;
```
clbartoli:
```suggestion
pragma solidity 0.7.0;
```

contracts/Request.sol
```diff
@@ -8,12 +8,14 @@ contract Request {
...
-  //
-  // The hash of the request is computed in the constructor to guarantee consistency. Otherwise there could be a
-  // mismatch and a data request could be resolved with the result of another.
+ /**
+  * @dev A `Request` is constructed around a `bytes memory` value containing a well-formed Witnet data request serialized
+  * using Protocol Buffers. However, we cannot verify its validity at this point. This implies that contracts using
+  * the WRB should not be considered trustless before a valid Proof-of-Inclusion has been posted for the requests.
+  * The hash of the request is computed in the constructor to guarantee consistency. Otherwise there could be a
+  * mismatch and a data request could be resolved with the result of another.
+  * @param _bytecode Witnet request in bytes
```
clbartoli:
```suggestion
  * @param _bytecode Witnet request in bytes.
```

contracts/SafeMath.sol
```diff
@@ -0,0 +1,151 @@
+pragma solidity ^0.6.0;
```
clbartoli: Not sure if the version should be variable since it is a library

contracts/UsingWitnet.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.0;
+pragma solidity >=0.5.3 <0.7.0;
```
clbartoli: Fixed version

contracts/UsingWitnetBytes.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.0;
+pragma solidity >=0.5.3 <0.7.0;
```
clbartoli: fixed version

contracts/Witnet.sol
```diff
@@ -1,10 +1,15 @@
-pragma solidity ^0.5.0;
+pragma solidity >=0.5.3 <0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "./BufferLib.sol";
 import "./CBOR.sol";
 
 
+/**
+ * @title A library for decoding Witnet request results
```
clbartoli:
```suggestion
 * @title A library for decoding Witnet request results.
```

contracts/Witnet.sol
```diff
@@ -1,10 +1,15 @@
...
+ * @notice The library exposes functions to check the Witnet request success
```
clbartoli:
```suggestion
 * @notice The library exposes functions to check the Witnet request success.
```

contracts/WitnetRequestsBoard.sol
```diff
@@ -112,7 +118,12 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     _;
   }
 
-  constructor (address _blockRelayAddress, uint8 _repFactor) public {
+ /**
+  * @notice Include an address to specify the Witnet Block Relay and a replication factor
```
clbartoli:
```suggestion
  * @notice Include an address to specify the Witnet Block Relay and a replication factor.
```

contracts/Request.sol
mariocao: This contract is used (inherited) by the truffle box, so we could allow a range of solidity versions. :)

contracts/UsingWitnet.sol
mariocao: True. :)

contracts/UsingWitnetBytes.sol
mariocao: Same as with Request.

contracts/Witnet.sol
mariocao: I left titles without points on purpose... don't know if it's the best or not... :frowning_man:

PR #73, opened 2020-04-08
Prevent loss of funds when using the UsingWitnet contract

shargon:
It has been verified that there is a different behaviour between `UsingWitnet` and `UsingWitnetBytes`. The first one, does not send the total of ether that has been sent during the transaction, instead it sends the sum of the two parameters of the fee. This behaviour can lead to the burning of funds if the contract that implements it, does not include a withdrawal mechanism.

It is recommended to unify the criteria of both contracts, so the `UsingWitnetBytes` mechanism, should be the standard one.

PS: This change requires modifying the examples `witnet/truffle-box` and `stampery-labs/witnet-pricefeed-example`.

### DIFF
--- a/contracts/UsingWitnet.sol
+++ b/contracts/UsingWitnet.sol
@@ -34,12 +34,11 @@ contract UsingWitnet {
   * @notice Send a new request to the Witnet network
   * @dev Call to `post_dr` function in the WitnetRequestsBoard contract
   * @param _request An instance of the `Request` contract
-  * @param _requestReward Reward specified for the user which posts the request into Witnet
-  * @param _resultReward Reward specified for the user which posts back the request result
+  * @param _tallyReward Reward specified for the user which post the Data Request result.
   * @return Sequencial identifier for the request included in the WitnetRequestsBoard
   */
-  function witnetPostRequest(Request _request, uint256 _requestReward, uint256 _resultReward) internal returns (uint256) {
-    return wrb.postDataRequest.value(_requestReward + _resultReward)(_request.bytecode(), _resultReward);
+  function witnetPostRequest(Request _request, uint256 _tallyReward) internal returns (uint256) {
+    return wrb.postDataRequest.value(msg.value)(_request.bytecode(), _tallyReward);
   }
 
   /**
--- a/test/helpers/UsingWitnetTestHelper.sol
+++ b/test/helpers/UsingWitnetTestHelper.sol
@@ -19,8 +19,8 @@ contract UsingWitnetTestHelper is UsingWitnet {
 
   constructor (address _wrbAddress) UsingWitnet(_wrbAddress) public { }
 
-  function _witnetPostRequest(Request _request, uint256 _requestReward, uint256 _tallyReward) external payable returns(uint256 id) {
-    return witnetPostRequest(_request, _requestReward, _tallyReward);
+  function _witnetPostRequest(Request _request, uint256 _tallyReward) external payable returns(uint256 id) {
+    return witnetPostRequest(_request, _tallyReward);
   }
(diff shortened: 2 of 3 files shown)


aesedepece:
Great point. We discussed this issue only a couples of weeks ago:

- **Option A**: as in `UsingWitnet`.
  - `requestReward` is explicitely stated when posting the request. Both rewards are known upfront.
  - The `witnetPostRequest` can only spend as much as `requestReward + tallyReward`.
  - If a user puts excess value, it remains in the consumer contract. It is therefore up to the user to provide a method for withdrawing or decide to "burn" it.

- **Option B**: as in `UsingWitnetBytes`.
  - `requestReward` is not explicitely stated when posting the request, but rather implicitly computed as `requestReward = value - tallyReward`.
  - The `witnetPostRequest` method drains up all available value.
  - If a user puts excess value, it goes to the relayer as it becomes part of `requestReward`.

Both options have clear pros and cons from the security and from the user experience standpoints.

I was more inclined towards option A, as the "phantom" implicit arguments in form of excess value from option B felt totally counter intuitive to me.

My main point was about "transaction value discovery", i.e. how does the caller of `witnetPostRequest` know what value needs to be put in the transaction? What happens the user miscalculates the total value?

In Option A, being both rewards explicitly stated, it becomes apparent that the value needs to equal the sum. Then if the user "underpays", a `require` somewhere reverts the transaction and the error message is eloquent about the shortage in value.

In Option B the transaction value still needs to be the sum of both rewards, but the `requestReward` is not implicitly stated. What I don't like about it is that on one hand we are totally concealing the existence of `requestReward`, which is the main reward, value-wise; and on the other hand, any value greater than `tallyReward` will make the transaction succeed, even if that means that `requestReward` will equal zero. There's no error message if the user miscalculates the total value — it will totally feel like everything went fine, but chances are that the request will never be relayed, as a too low `requestReward` would make a bridge operate at a loss.

As of today, I must say I'm not totally satisfied with either option. I still think that rewards should be explicit so that undervalued requests can be detected upon posting, like in option A. But I'm obviously not comfortable with the idea of forcing users to decide whether to implement an excess value withdrawal method into the consumer contract, or otherwise take the risk of burning ETH by accident (option B is no better in that sense, although instead of burning you are overpaying the relayer).

My feeling right now is that with regards to `UsingWitnet`, we should stick with something similar to option A, but with an additional `require` for preventing the value of the transaction from exceeding the sum of the rewards, so that no excess value can remain in the consumer contract.

Beyond `UsingWitnet`, at the `WitnetRequestBoard` level, sincerely, anything works.

aesedepece:
Of course each of us devs have our own trade-off considerations with regards to this sort of dichotomy, but I think we can find a middle ground that addresses most of everyone's concerns.

0d4rujd:
Maybe there is an option C.

Set a percentage: The user sets the quantity in msg.value and the parameter defines the % that goes to each part. This value should't be exceeded, since the function is not intended to deposit ethereum. In case the contract wants to deposit ethereum, I think you should use another payable function, but not that one.

girazoki:
I think, as @aesedepece said, the way to go is to insert a require in UsingWitnet checking whether msg.value exceeds _requestReward + _tallyReward. This for me is the most transparent yet secure option to prevent burning of funds in neither direction

mariocao:
Reading all above comments, I like option A, i.e. explicitly stating `requestReward` and `tallyReward` but making sure that there are not more funds that both rewards by using a `require` statement.

mariocao:
@shargon if you agree, do you mind amending the Pull Request with the changes? :smile:

mariocao:
Close by #86

PR #77, opened 2020-04-14
Implement only ABS members can report results

### LINKED ISSUE #74: Race Condition: Transaction Order Dependence in WitnetRequestBoard
Finding discovered during [Red4Sec](https://red4sec.com) security audit.

## Description

There is a Transaction Order Dependence vulnerability in the WitnetRequestBoard contract leading to a race condition. 

The reportResult function implements numerous entries to verify the TallyPoi, however, none of them is related to the sender of the transaction, therefore, anyone who issues a transaction with such information to the blockchain, will be the first to execute its logic. Furthermore, on [line:245](https://github.com/witnet/witnet-ethereum-bridge/blob/9d1f3be651ac644056303d18c20eecc4fc490472/contracts/WitnetRequestsBoard.sol#L245
) of WitnetRequestBoard.sol file, funds are sent to the sender's account, which allows obtaining an economic benefit (acquiring the response fee) by exploiting the vulnerability.

In the following image, we can see that there is no verification to confirm that the sender is legitimate.

v001_1

&nbsp;

## Impact

Allows an attacker to obtain the fee reward on a recurring basis without doing any work.

&nbsp;
## Steps to reproduce

In order to exploit this vulnerability, we have developed a tool that listens for all the existing transactions in the memory pool. As soon as it detects that a transaction has been received with the following characteristics;

- Destination:  0x72cf2dff1cf373ad818b37d8196c9d372a8f2ade (WitnessRequestBoard - WRB).
- MethodId: 0x8a68f33e (reportResult).

it will automatically generate an identical transaction (copying all the values ​​from the original) but using a higher gas price. 

In this way we prioritize our transaction and ensure that it has more possibilities to enter before the original bridge transaction, thus receiving the result fee without having done any work. 

v001_2

Next, we can see how the attacker's transaction has copied all the input values ​​of the original, except for the gasPrice that has been multiplied by 2. This allows us to gain priority and therefore our transaction will be entered before the original transaction despite being sent later.

https://goerli.etherscan.io/tx/0xdfb86100dcf4d22533d6b7a813770b8ab793d4774c375b72211d79e3a03c2799
In the following link we can check how the original transaction, despite entering the same block (2489093), is processed after the attacker's, which returns the following error: "Result already included".

https://goerli.etherscan.io/tx/0xa533545c9b930fabffb37ba466b19e45ca4ac37ed8f452633fd7880b85f1d724

v001_3

&nbsp;
## Code References 

https://github.com/witnet/witnet-ethereum-bridge/blob/9d1f3be651ac644056303d18c20eecc4fc490472/contracts/WitnetRequestsBoard.sol#L244
https://github.com/witnet/witnet-ethereum-bridge/blob/9d1f3be651ac644056303d18c20eecc4fc490472/contracts/WitnetRequestsBoard.sol#L245

&nbsp;
## Mitigation
To avoid this type of attack, it is necessary that the PoI validation include the sender's address to avoid this type of attack.

clbartoli:
This PR implements the absMember modifier in WitnetRequestBoard.sol when calling reportResult. 
Now only the members of the active bridge set are able to report results, before there were no barriers to call this function.

The tests wrb.js, using_witnet.js and using_witnet_bytes.js have been adapted to the new modifier.

Closes #74

### DIFF
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -118,6 +118,11 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
       "Not a valid VRF");
     _;
   }
+  // Ensures the address is from a member of the active bridge set
+  modifier absMember(address _address) {
+    require(abs.absMembership(_address) == true, "Not a member of the ABS");
+    _;
+  }
 
  /**
   * @notice Include an address to specify the Witnet Block Relay and a replication factor.
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -118,7 +118,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
       "Not a valid VRF");
     _;
   }
-  // Ensures the address is from a member of the active bridge set
+  // Ensures the address belongs to the active bridge set
   modifier absMember(address _address) {
     require(abs.absMembership(_address) == true, "Not a member of the ABS");
     _;
@@ -235,9 +235,9 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     uint256 _epoch,
     bytes calldata _result)
     external
-    absMember(msg.sender)
     drIncluded(_id)
     resultNotIncluded(_id)
+    absMember(msg.sender)
  {
 
     // Ensures the result byes do not have zero length
@@ -256,8 +256,6 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     requests[_id].result = _result;
     msg.sender.transfer(requests[_id].tallyReward);
 
-    // Push msg.sender to abs
-    //abs.pushActivity(msg.sender, block.number);
     emit PostedResult(msg.sender, _id);
   }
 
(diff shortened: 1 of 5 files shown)


aesedepece:
I understand that the bottom line in this is restricting which Ethereum identities can interact with `reportResult()` so that only actual bridge nodes that have proven their liveness can front-run each other by means of gas price outbidding (which is allowed by design).

contracts/WitnetRequestsBoard.sol
```diff
@@ -245,7 +251,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     msg.sender.transfer(requests[_id].tallyReward);
 
     // Push msg.sender to abs
-    abs.pushActivity(msg.sender, block.number);
+    //abs.pushActivity(msg.sender, block.number);
```
girazoki, line 254: Remove comment

contracts/WitnetRequestsBoard.sol
```diff
@@ -117,6 +117,11 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
       "Not a valid VRF");
     _;
   }
+  // Ensures the address is from a member of the active bridge set
```
girazoki, line 120:
```suggestion
  // Ensures the address belongs to the active bridge set
```

PR #79, opened 2020-04-15
fix: check posted result length

### LINKED ISSUE #75: Drain of Contract Funds that could trigger a Denial of Service
Finding discovered during Red4Sec security audit.

## Description
Reviewing _**WitnetRequestBoard**_ smart contract it was detected that the reportResult method incorrectly implements the logic necessary to verify whether a result has been previously included or not. 
&nbsp;

## Impact
This failure could tentatively  cause a drain on the contract funds by any attacker and produce a denial of service.
&nbsp;

## Steps to reproduce
As we can see in the following screenshot, there is a modifier called resultNotIncluded, which uses the size of the result to check whether or not it has been previously included. It should be noted that a result can also be empty ("").

image

&nbsp;
In the **_CBOR_** library it is checked that the size is greater than 0 to consider it a valid CBOR, so the modifier that has been observed in the previous image, apparently would turn out to be valid.

image
&nbsp;

However, as can be seen in the following image, it is not verified that the result is a valid CBOR. This verification is only done during reading and not during reporting. This would allow us to bypass the _resultNotIncluded_ modifier in case an empty result ("") is sent.
In this way, it would be possible to collect the fee for said result without marking it as consumed, being able to send the same transaction recursively and collecting the same fee from the contract funds and not from the part reserved for the bridge that serves it.

In this way, the attacker could completely drain the funds from the _WitnetRequestBoard_ contract, in turn making a _Denial of Service_ (DoS), since the bridges could not receive the fee they had intended to carry out their work and therefore the execution of their transactions would fail.

image

&nbsp;
## Code References
https://github.com/witnet/witnet-ethereum-bridge/blob/9d1f3be651ac644056303d18c20eecc4fc490472/contracts/WitnetRequestsBoard.sol#L101

https://github.com/witnet/witnet-ethereum-bridge/blob/9d1f3be651ac644056303d18c20eecc4fc490472/contracts/WitnetRequestsBoard.sol#L224
&nbsp;

## Mitigation
It is recommended to add a boolean type field in the DataRequest struct as a flag to check whether the result has already been reported or not.

girazoki:
this PR checks the length of the posted result and reverts if it is zero. This avoids potential re-entrancy attacks if the block relay accepts the PoI.

This PR closes #75

### DIFF
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -101,6 +101,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     require(requests[_id].result.length == 0, "Result already included");
     _;
   }
+
 // Ensures the VRF is valid
   modifier vrfValid(
     uint256[4] memory _poe,
@@ -232,6 +233,11 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     drIncluded(_id)
     resultNotIncluded(_id)
  {
+
+    // Ensures the result byes do not have zero length
+    // This would not be a valid encoding with CBOR and could trigger a reentrancy attack
+    require(_result.length != 0, "Result has zero length");
+
     // this should leave it ready for PoI
     uint256 resHash = uint256(sha256(abi.encodePacked(requests[_id].drHash, _result)));
     require(
--- a/test/wrb.js
+++ b/test/wrb.js
@@ -669,6 +669,74 @@ contract("WRB", accounts => {
         "Result already included"
       )
     })
+    it("should revert because of insertion of empty result",
+      async () => {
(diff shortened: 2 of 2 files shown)


contracts/WitnetRequestsBoard.sol
```diff
@@ -232,6 +233,11 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     drIncluded(_id)
     resultNotIncluded(_id)
  {
+
+    // Ensures the result byes do not have zero length
+    // This would not a valid encoding with CBOR and could trigger a reentrancy attack
```
clbartoli, line 238:
```suggestion
    // This would not be a valid encoding with CBOR and could trigger a reentrancy attack
```

PR #86, opened 2020-04-27
fix: preventing burning user funds

mariocao:
This PR prevents burning user funds due to non-existent checks in post request and upgrade request  functions, located in both `UsingWitnet` and `WitnetRequestsBoard` contracts.

To discuss what to do with `UsingWitnetBytes` as it has not been properly maintained:

- lacks of some modifiers (e.g. `witnetRequestAccepted`)
- deviates from function signatures used in `UsingWitnet.sol`

Supersedes #73

### DIFF
--- a/contracts/UsingWitnet.sol
+++ b/contracts/UsingWitnet.sol
@@ -30,6 +30,13 @@ contract UsingWitnet {
     _;
   }
 
+  // // Ensures that user-specified rewards are equal to the total transaction value to prevent users from burning any excess value
+  modifier rewardsValid(uint256 _requestReward, uint256 _resultReward) {
+    require(_requestReward + _resultReward >= _requestReward, "The sum of rewards overflows");
+    require(msg.value == _requestReward + _resultReward, "The sum of rewards should be equal to transaction value");
+    _;
+  }
+
   /**
   * @notice Send a new request to the Witnet network
   * @dev Call to `post_dr` function in the WitnetRequestsBoard contract
@@ -38,7 +45,11 @@ contract UsingWitnet {
   * @param _resultReward Reward specified for the user which posts back the request result
   * @return Sequencial identifier for the request included in the WitnetRequestsBoard
   */
-  function witnetPostRequest(Request _request, uint256 _requestReward, uint256 _resultReward) internal returns (uint256) {
+  function witnetPostRequest(Request _request, uint256 _requestReward, uint256 _resultReward)
+    internal
+    rewardsValid(_requestReward, _resultReward)
+  returns (uint256)
+  {
     return wrb.postDataRequest.value(_requestReward + _resultReward)(_request.bytecode(), _resultReward);
   }
 
@@ -61,10 +72,14 @@ contract UsingWitnet {
   * @notice Upgrade the rewards for a Data Request previously included.
   * @dev Call to `upgrade_dr` function in the WitnetRequestsBoard contract.
   * @param _id The sequential identifier of a request that has been previously sent to the WitnetRequestsBoard.
-  * @param _tallyReward Reward specified for the user which post the Data Request result.
+  * @param _requestReward Reward specified for the user which posts the request into Witnet
+  * @param _resultReward Reward specified for the user which post the Data Request result.
   */
-  function witnetUpgradeRequest(uint256 _id, uint256 _tallyReward) internal {
-    wrb.upgradeDataRequest.value(msg.value)(_id, _tallyReward);
+  function witnetUpgradeRequest(uint256 _id, uint256 _requestReward, uint256 _resultReward)
+    internal
+    rewardsValid(_requestReward, _resultReward)
+  {
+    wrb.upgradeDataRequest.value(msg.value)(_id, _resultReward);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


contracts/UsingWitnet.sol
```diff
@@ -30,6 +30,13 @@ contract UsingWitnet {
     _;
   }
 
+  // Ensures that user-specified rewards are equal to the total transatraction amount, thus avoiding burning user funds
```
aesedepece, line 33:
```suggestion
  // Ensures that user-specified rewards are equal to the total transaction value to prevent users from burning any excess value
```

contracts/WitnetRequestsBoard.sol
```diff
@@ -163,9 +163,23 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
...
+    override
+  {
+    if (requests[_id].drHash != 0) {
+      require(
+        msg.value == _tallyReward,
+        "Result reward should be equal to txn value (data request was included)"
+      );
+      requests[_id].tallyReward = SafeMath.add(requests[_id].tallyReward, _tallyReward);
+    } else {
+      requests[_id].inclusionReward = SafeMath.add(requests[_id].inclusionReward, SafeMath.sub(msg.value, _tallyReward));
```
aesedepece, line 180: Seeing `SafeMath.sub` here makes me wonder whether there's a test for asserting that this function reverts when `_tallyReward < msg.value`.

contracts/WitnetRequestsBoard.sol
```diff
@@ -163,9 +163,23 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
...
+    external
+    payable
+    payingEnough(msg.value, _tallyReward)
+    resultNotIncluded(_id)
+    override
+  {
+    if (requests[_id].drHash != 0) {
+      require(
+        msg.value == _tallyReward,
+        "Result reward should be equal to txn value (data request was included)"
```
aesedepece, line 176:
```suggestion
        "Result reward should equal txn value because request reward can't be updated for requests that have already been included"
```

contracts/WitnetRequestsBoard.sol
```diff
@@ -163,9 +163,23 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
   /// @dev Increments the rewards of a data request by adding more value to it. The new request reward will be increased by msg.value minus the difference between the former tally reward and the new tally reward.
   /// @param _id The unique identifier of the data request.
   /// @param _tallyReward The new tally reward. Needs to be equal or greater than the former tally reward.
-  function upgradeDataRequest(uint256 _id, uint256 _tallyReward) external payable payingEnough(msg.value, _tallyReward) override {
-    requests[_id].inclusionReward = SafeMath.add(requests[_id].inclusionReward, SafeMath.sub(msg.value, _tallyReward));
-    requests[_id].tallyReward = SafeMath.add(requests[_id].tallyReward, _tallyReward);
+  function upgradeDataRequest(uint256 _id, uint256 _tallyReward)
```
aesedepece, line 166: I think we should stick to either on of `tallyReward` or `resultReward`. The dual naming is confusing even for myself.

contracts/WitnetRequestsBoard.sol
mariocao: The reason why this `require` message is not longer is because of the linter.

```
contracts/WitnetRequestsBoard.sol
  174:6    warning    Error message exceeds max length of 76 characters    error-reason
```

contracts/WitnetRequestsBoard.sol
mariocao: It should not revert because `_tallyReward < msg.value` is the expected behaviour.

However, the `SafeMath.sub` there is not really required as the `payingEnough` modifier is checking that `msg.value > _tallyReward`.

contracts/WitnetRequestsBoard.sol
mariocao: The same happens with `inclusionReward` and  `requestReward`. I agree with this refactor but maybe it should be done in another PR to modified both cases at the same time.

contracts/WitnetRequestsBoard.sol
mariocao: They added splitting literals in here: https://github.com/ethereum/solidity/pull/7524

mariocao: But linter complains!!! :rofl:

contracts/WitnetRequestsBoard.sol
```diff
@@ -163,9 +163,23 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
...
+    external
+    payable
+    payingEnough(msg.value, _tallyReward)
+    resultNotIncluded(_id)
+    override
+  {
+    if (requests[_id].drHash != 0) {
+      require(
+        msg.value == _tallyReward,
+        "Result reward should equal txn value (request reward already given)"
```
aesedepece, line 176:
```suggestion
        "Transaction value should equal result reward argument (as request reward was already paid to relayer)"
```

mariocao: Remember that the linter complains:

```
Error message exceeds max length of 76 characters    error-reason
```

mariocao: Proposal: `Txn value should equal result reward argument (request reward already paid)`

PR #90, opened 2020-05-06
Add epoch condition in wrb to avoid replay poi attack

### LINKED ISSUE #88: Replay PoI Attack
Finding discovered during [Red4Sec](https://red4sec.com) security audit.

## Description
A Replay Attack is a kind of network attack that occurs when a cybercriminal eavesdrop on a secure network communication, intercepts it and maliciously or fraudulently delays or resends it to misdirect the receiver into carrying out unexpected behaviours, in order to get any kind of profit from this action.  The main objective of a Replay attack consists in sending previously sent data in a malicious manner, this can be considered as a lower tier version of a "Man-in-the-middle attack" as the message is intercepted and replayed by an attacker to impersonate the original sender.

If we extrapolate this to the blockchain, the interception process becomes easier since the information in Ethereum’s blockchain is public, it is the smart contract’s duty to implement the measures in order to avoid this type of attack. 
&nbsp;

## Impact
Red4Sec has detected a Replay Attack vulnerability in the **_reportDataRequestsInclusion_** and **_reportResult_** methods.  Such vulnerability will allow the attacker the following:
- To take economically profit
- To reuse past or outdated values and be able to validate them in the present time.

For example, in a scenario where a decentralized exchange uses Witnet’s oracles to obtain the current bitcoin price, a malicious user would be capable of sending the bitcoin price set 2 years ago and it would be verified and validated to the current date.
&nbsp;

## Steps to reproduce
In the following image, which belongs to the **_reportResult_** method, you can observe that the identifier argument of the request (**__id_**) is not used to verify the _TallyPoI_ input arguments which can always be used as long as they are valid. 
In addition to this, it uses the _resHash_ value which is formed by the result (__result_) and the _drHash_.

In fact, the _drHash_ it’s the only value that could avoid a replay attack in advance, if all the previous values are validated beforehand, upcoming we will be able to confirm how such Hash is formed.

image

In the following image you can corroborate that the _drHash_ it’s formed uniquely by the first value of the PoI (**__poi[0]_**) and a hash from the **_dataRequest_** (**_drOutputHash_**) which means there is no correlation with the identifier of the request. Furthermore, we can observe that in order to verify the PoI (_**verifyDrPoi**_) you don’t need the identifier, in such manner we would be capable of obtaining the same _drHash_ from a past transaction, as long as we reuse their values, this means that there is a possibility to use a different identifier if we reuse the past values.

Even though we will be able to replicate the **_reportDataRequestsInclusion_**, the reward of this transaction will go to the original bridge, however, our objective is to obtain the same _drHash_ with the purpose of falsifying the result and obtaining the next reward (_**reportResult**_).

image

In conclusion, and as disclosed in the premises detailed above, the contract is vulnerable to a Replay Attack, which will allow us to return a past value and validate it by the oracle as a present value, in addition to obtaining a reward.

## Proof of Concept
The lifecycle of a Witnet request consists in 4 transactions. On the user’s side the request (U1), and on the bridge’s side the **_claimDataRequests_** (C1), the **_reportDataRequestInclusion_** (I1) and the **_reportResult_** (R1).

Since the target of this attack is to validate a value of the past in the present, we will need to save the values of these transactions in order to configure our attack.

In order to successfully execute our attack, we must configure the exploit beforehand. This exploit is listening to the mempool for any received transaction in the contract **0xe361333f6a4c441381c37df3474fd9eb46a6e610** (WRB) so that it executes only when it receives the claimDataRequest of our transaction (C2) by using the following condition: **^0xc9bcae14([a-fA-F0-9]+)0{30}140{61}[a-fA-F0-9]+$** 

image

In order to execute it we will have to send two transactions with less priority than C2 since both transactions will be the **_reportDataRequestInclusion_** (I2) and the **_reportResult_** (R2) which must necessarily get into the blockchain after the **_claimDataRequests_** (C2), this is the reason we use a gas price lower to the one in the transaction C2 (0.9 of C2). 

The input of the transaction I2 should by exactly as I1, changing the I1 identifier for our identifier. The same process must be performed on the second transaction, our **_reportResult_** (R2), in order to achieve this, we will reuse the R1 values by modifying only the necessary identifier. 

image

After the environment is properly configured, we must only wait for the reception of the user’s request (U2) so our exploit proceeds to inject both transactions.

image

In the following image you can observe that the _**reportDataInclusion**_ (one of our injected tx) appears in the same block as the **_claimDataRequests_** (one of our injected tx) and the false result (**_reportResult_**) is validated in the next block.

image

Both transactions have been successfully processed by the WRB, which means that in addition to faking the result by using values of an older PoI, we will receive the correspondent fee to **_reportResult_**.

image

The following transactions were used to carry out the PoC:
- U1: https://goerli.etherscan.io/tx/0x1ac4c2d1e5ac4ee619ba31e26f739038f35c6967eb13b3abf900777301d6fa13
- C1: https://goerli.etherscan.io/tx/0xb3e9c0e07f04691650fedb46ffba11ace82bce7dd9d9412596782d1de99e40ea
- I1: https://goerli.etherscan.io/tx/0xc20cd749a94f16ebdf6098dcab0425e1b5afafd3e24738226f68059bbc54cd2e
- R1: https://goerli.etherscan.io/tx/0x358989567b2190ce7b79d36554495f69c36b4125b39a1e34fa02fe6b5c41bafb
 
- I2:  https://goerli.etherscan.io/tx/0x285f387f06c6bc68eec4c717de0acc206d89dc38a8a091cc8a64ede5f1473faa
- R2: https://goerli.etherscan.io/tx/0x8e626bb6ad8c83f24f3c336d60dfbad8f869a1652392e0db291f2aa23e665ca9

&nbsp;

## Code Reference
https://github.com/witnet/witnet-ethereum-bridge/blob/cf4ffe9434be79ec29079a3fcaf885adfb213cb8/contracts/WitnetRequestsBoard.sol#L203
https://github.com/witnet/witnet-ethereum-bridge/blob/cf4ffe9434be79ec29079a3fcaf885adfb213cb8/contracts/WitnetRequestsBoard.sol#L251
&nbsp;

## Mitigation
- Implement protections against replay attacks.
- Add the request identifier to the _drHash_.

image

clbartoli:
This PR solves the problem found by Red4sec #88.  The idea is to check the epoch when reporting a data request, reporting the data request inclusion and reporting the result. It is required that each of the just mentioned functions are executed chronologically.

This PR adds:
- epoch field in DataRequest struct
- require in reportDataRequestInclusion
- require in reportResult
- tests in wrb.js

Fix #88

### DIFF
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -30,6 +30,8 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     bytes result;
     // Block number at which the DR was claimed for the last time
     uint256 blockNumber;
+    // The epoch of the blockHash in which the last transaction related to the dr has been included
+    uint256 epoch;
     uint256 drHash;
     address payable pkhClaim;
   }
@@ -162,6 +164,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     request.dr = _serialized;
     request.inclusionReward = SafeMath.sub(msg.value, _tallyReward);
     request.tallyReward = _tallyReward;
+    request.epoch = blockRelay.getLastEpoch();
 
     // Push the new request into the contract state
     requests.push(request);
@@ -228,6 +231,14 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
  {
     // Check the data request has been claimed
     require(dataRequestCanBeClaimed(requests[_id]) == false, "Data Request has not yet been claimed");
+
+    // Ensures the request inclusion is reported after the epoch in which the request was posted
+    require(
+      requests[_id].epoch < _epoch,
+      "the dr inclusion must be reported after the dr is posted into the WRB");
+    // Update the dr epoch
+    requests[_id].epoch = _epoch;
+
     uint256 drOutputHash = uint256(sha256(requests[_id].dr));
     uint256 drHash = uint256(sha256(abi.encodePacked(drOutputHash, _poi[0])));
 
@@ -265,6 +276,8 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     resultNotIncluded(_id)
     absMember(msg.sender)
  {
+    // Ensures the result was published in a later block than the request
+    require(requests[_id].epoch <= _epoch, "the result must be reported after the request is included");
 
     // Ensures the result byes do not have zero length
     // This would not be a valid encoding with CBOR and could trigger a reentrancy attack
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -236,7 +236,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     // Ensures the request inclusion is reported after the epoch in which the request was posted
     require(
       requests[_id].epoch < _epoch,
-      "the dr inclusion must be reported after the dr is posted into the WRB");
+      "The request inclusion must be reported after it is posted into the WRB");
     // Update the dr epoch
     requests[_id].epoch = _epoch;
 
@@ -278,7 +278,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     absMember(msg.sender)
  {
     // Ensures the result was published in a later block than the request
-    require(requests[_id].epoch <= _epoch, "the result must be reported after the request is included");
+    require(requests[_id].epoch <= _epoch, "The result cannot be reported before the request is included");
     // Update epoch of the request
     requests[_id].epoch = _epoch;
 
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -30,7 +30,8 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     bytes result;
     // Block number at which the DR was claimed for the last time
     uint256 blockNumber;
-    // The epoch of the blockHash in which the last transaction related to the dr has been included
+    // The epoch of the block including the last transaction related to the dr
+    // (postDataRequest, reportDataRequestInclusion, reportResult)
     uint256 epoch;
     uint256 drHash;
     address payable pkhClaim;
@@ -278,6 +279,8 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
  {
     // Ensures the result was published in a later block than the request
     require(requests[_id].epoch <= _epoch, "the result must be reported after the request is included");
+    // Update epoch of the request
+    requests[_id].epoch = _epoch;
 
     // Ensures the result byes do not have zero length
     // This would not be a valid encoding with CBOR and could trigger a reentrancy attack
--- a/test/wrb.js
+++ b/test/wrb.js
@@ -511,7 +511,6 @@ contract("WRB", accounts => {
       const resBytes = web3.utils.fromAscii("This is a result")
       const data1 = "0x" + sha.sha256(web3.utils.hexToBytes(drBytes))
       var dummySybling = 1
-      const epoch = 0
 
       // VRF params
       const publicKey = [data.publicKey.x, data.publicKey.y]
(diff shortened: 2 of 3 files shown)


contracts/WitnetRequestsBoard.sol
```diff
@@ -204,6 +206,12 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     external
     drNotIncluded(_id)
  {
+    // Ensures the dr.epoch is smaller than the last epoch reported to the block relay
+    uint256 witnetEpoch = blockRelay.getLastEpoch();
+    require(requests[_id].epoch < witnetEpoch, "the current epoch is not compatible with the dr epoch");
```
aesedepece, line 211: Error messages should ideally be explicit about what the user did wrong.

contracts/WitnetRequestsBoard.sol
```diff
@@ -239,6 +247,10 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     resultNotIncluded(_id)
     absMember(msg.sender)
  {
+    // Ensures the epoch when the dr inclusion was reported is smaller or equal
+    // the last epoch included in the block relay
+    uint256 witnetEpoch = blockRelay.getLastEpoch();
+    require(requests[_id].epoch <= witnetEpoch, "the current epoch is not compatible with the dr epoch");
```
aesedepece, line 253: I think that rather than checking here if "the request was published in the past" we should check that "the result was published in a later block than the request".

contracts/WitnetRequestsBoard.sol
aesedepece, line 211: What's this check for? I think the `witnetEpoch` can never be lower than `requests[_id].epoch` and therefore this requirement is always met unless both terms are equal. I wonder what we are trying to prevent here.

clbartoli: Yes, my fault. The check should be between the `dr.epoch` and the input `_epoch`. It's just been changed.

contracts/WitnetRequestsBoard.sol
```diff
@@ -27,6 +27,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     uint256 tallyReward;
     bytes result;
     uint256 timestamp;
+    uint256 epoch;
     uint256 drHash;
```
girazoki, line 31: Lets try to comment what epoch is, as it might not be straigh forward.

test/wrb.js
```diff
@@ -540,17 +540,84 @@ contract("WRB", accounts => {
...
       await waitForHash(tx3)
 
       // assert it fails when trying to report the dr inclusion again
-      await truffleAssert.reverts(wrbInstance.reportDataRequestInclusion(id1, [dummySybling], 1, blockHeader2, epoch, {
+      await truffleAssert.reverts(wrbInstance.reportDataRequestInclusion(id1, [dummySybling], 1, blockHeader2, epoch1, {
         from: accounts[1],
       }), "DR already included")
     })
 
+    it("should revert when trying to report a dr inclusion before the epoch in which the is posted ", async () => {
```
girazoki, line 554:
```suggestion
    it("should revert when trying to prove inclusion of a DR in an epoch inferior than the one in which DR was posted and when trying to prove a result in an epoch inferior than the one in which the DR was included ", async () => {
```

girazoki:
Please check why builds are not passing

contracts/WitnetRequestsBoard.sol
```diff
@@ -27,6 +27,8 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     uint256 tallyReward;
     bytes result;
     uint256 timestamp;
+    // The epoch of the blockHash in which the last transaction related to the dr has been included
```
mariocao, line 30: If I understood it correctly, the comment could also mention that this value has different meanings in the DR lifecycle:
1. when dr is posted
2. when dr transaction is included

And then, (2) is used to check that the dr result happened after :)

contracts/WitnetRequestsBoard.sol
```diff
@@ -204,6 +206,11 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     external
     drNotIncluded(_id)
  {
+    // Ensures the request inclusion is reported after the epoch in which the request was posted
+    require(requests[_id].epoch < _epoch, "the epoch when the request was posted has not passed");
```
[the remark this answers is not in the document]
mariocao: I fear that this error message is bigger than 76 chars :/

contracts/WitnetRequestsBoard.sol
```diff
@@ -222,6 +226,13 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     external
     drNotIncluded(_id)
  {
+    // Ensures the request inclusion is reported after the epoch in which the request was posted
+    require(
+      requests[_id].epoch < _epoch,
+      "the dr inclusion must be reported after the dr is posted into the WRB");
```
mariocao, line 232:
```suggestion
      "The dr inclusion must be reported after the dr is posted into the WRB");
```

PR #92, opened 2020-05-08
fix: read modifier dependent state before external calls

girazoki:
This PR modifies the state upon which `reportDrInclusion` and `reportResult` depend before the externals calls are made

### DIFF
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -224,6 +224,9 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
  {
     uint256 drOutputHash = uint256(sha256(requests[_id].dr));
     uint256 drHash = uint256(sha256(abi.encodePacked(drOutputHash, _poi[0])));
+
+    // Update the state upon which this function depends before the external call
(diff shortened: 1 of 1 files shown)


contracts/WitnetRequestsBoard.sol
```diff
@@ -220,14 +220,16 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
  {
     uint256 drOutputHash = uint256(sha256(requests[_id].dr));
     uint256 drHash = uint256(sha256(abi.encodePacked(drOutputHash, _poi[0])));
+
+    // update the state upon which this function depends before the external call
```
mariocao, line 224: Just following the same format as previous lines (first letter is capital) :)
```suggestion
    // Update the state upon which this function depends before the external call
```

contracts/WitnetRequestsBoard.sol
```diff
@@ -260,14 +262,15 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
 
     // this should leave it ready for PoI
```
mariocao, line 263: Maybe replace this line with the comment you just inserted

PR #97, opened 2020-05-08
fix: add non-existent checks in libraries and WRB

girazoki:
This PR adds some checks that were not existent in the smart contracts. Some of which have not been fixed:

Witnet-ethereum-block-relay\contracts\CentralizedBlockRelay.sol:156-157
In flushSlot() function, it is not verified that values are the expected. It should be verified -> **Non-existent (probably a mistake in the report)** 

Witnet-ethereum-bridge\contracts\ActiveBridgeSetLib.sol:50-51
In flushSlot() function, it is not verified that values are the expected. It should be verified
that the epoch is the consecutive or the same as the last inserted. -> **Verified in UpdateABS from WRB**

• Witnet-ethereum-bridge\contracts\WitnetRequestsBoardProxy.sol:120-125
When exiting for loop, instead of returning the default values. We recommend including a
revert() in order to avoid returning non-existent "controllers". -> **Default controllers can never be returned according to the algorithm (always an existent controller is returned)**

• Witnet-ethereum-bridge\contracts\WitnetRequestsBoard.sol:269-270
Witnet-ethereum-bridge\contracts\UsingWitnet.sol:66-67 -> **Fixed by #86**

• Witnet-ethereum-bridge\contracts\Witnet.sol:418-424
This function converts an array to a string but does not work correctly when the size is
greater than 3 digits. It should be verified and reverted when it exceeds this limit. -> **By definition the function fails if a string with more than 3 digits is inserted. This is expected behavior, and we do not pretend such a function to revert if more than 3 digits are inserted as that would prevent us from reading useful errors from Witnet.

### DIFF
--- a/.soliumignore
+++ b/.soliumignore
@@ -1,3 +1,4 @@
 node_modules
 contracts/Migrations.sol
 contracts/flattened/Flattened.sol
+test/TestBuffer.sol
--- a/contracts/BufferLib.sol
+++ b/contracts/BufferLib.sol
@@ -14,6 +14,12 @@ library BufferLib {
     uint32 cursor;
   }
 
+  // Ensures we access an existing index in an array
+  modifier notOutOfBounds(uint32 index, uint256 length) {
+    require(index < length, "Tried to read from a consumed Buffer (must rewind it first)");
+    _;
+  }
+
   /**
   * @notice Read and consume a certain amount of bytes from the buffer.
   * @param _buffer An instance of `BufferLib.Buffer`.
@@ -51,7 +57,7 @@ library BufferLib {
   * @param _buffer An instance of `BufferLib.Buffer`.
   * @return The next byte in the buffer counting from the cursor position.
   */
-  function next(Buffer memory _buffer) internal pure returns (byte) {
+  function next(Buffer memory _buffer) internal pure notOutOfBounds(_buffer.cursor, _buffer.data.length) returns (byte) {
     // Return the byte at the position marked by the cursor and advance the cursor all at once
     return _buffer.data[_buffer.cursor++];
   }
@@ -100,7 +106,7 @@ library BufferLib {
   * @param _buffer An instance of `BufferLib.Buffer`.
   * @return The `uint8` value of the next byte in the buffer counting from the cursor position.
   */
-  function readUint8(Buffer memory _buffer) internal pure returns (uint8) {
+  function readUint8(Buffer memory _buffer) internal pure notOutOfBounds(_buffer.cursor, _buffer.data.length) returns (uint8) {
     bytes memory bytesValue = _buffer.data;
     uint32 offset = _buffer.cursor;
     uint8 value;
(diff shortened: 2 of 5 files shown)


test/TestBuffer.sol
```diff
@@ -1,9 +1,9 @@
 pragma solidity 0.6.4;
+pragma experimental ABIEncoderV2;
 
 import "truffle/Assert.sol";
 import "../contracts/BufferLib.sol";
 
-
 contract TestBuffer {
```
mariocao, line 7:
```suggestion

contract TestBuffer {
```

contracts/BufferLib.sol
```diff
@@ -12,6 +12,12 @@ library BufferLib {
     uint64 cursor;
   }
 
+  // Ensures we access an existing index in an array
+  modifier notOutOfBounds(uint256 index, uint256 length) {
```
aesedepece, line 16: I'd say that for the scope of `BufferLib`, the `index` argument could use `uint32`.

contracts/BufferLib.sol
```diff
@@ -12,6 +12,12 @@ library BufferLib {
...
+    require(index < length, "Array index out of bounds");
```
aesedepece, line 17: I'd be surprised if a function called `readUint8` would fail with "array index out of bounds", given that neither the name or arguments of the function mention an array nor an index.

Maybe something like this? 
```suggestion
    require(index < length, "Tried to read from a consumed Buffer (must rewind it first)");
```

contracts/Witnet.sol
```diff
@@ -407,6 +407,7 @@ library Witnet {
    * @return The `string` representing its decimal value.
    */
   function utoa(uint64 _u) private pure returns(string memory) {
+    require(_u < 1000, "error, the library does not handle more than 3 digits");
```
aesedepece, line 410: This change actually breaks the behavior of the function as documented above:
```ts
/**
   * @notice Convert a `uint64` into a 1, 2 or 3 characters long `string` representing its
   * three less significant decimal values.
   * @param _u A `uint64` value.
   * @return The `string` representing its decimal value.
   */
```

.soliumignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 contracts/Migrations.sol
-contracts/flattened/Flattened.sol
\ No newline at end of file
+contracts/flattened/Flattened.sol
+test/TestBuffer.sol
```
aesedepece, line 4: Is this because of usage of `fallback`?

test/TestBuffer.sol
```diff
@@ -74,4 +75,126 @@ contract TestBuffer {
     uint64 actualBig = buf.readUint64();
     Assert.equal(uint(actualBig), uint(big), "Read multiple data chunks from the same Buffer (inner cursor works as expected)");
   }
+
+  function testReadUint8asUint16() external {
+    ThrowProxy throwProxy = new ThrowProxy(address(this));
+    bool r;
+    bytes memory error;
+    uint8 expected = 0xAA;
```
aesedepece, line 83: We are actually expecting this to fail, so I guess the values defined here are more `input` or something similar rather than `expected`.

.soliumignore
girazoki: yep

contracts/BufferLib.sol
girazoki: the problem is buffer.data.length is uint256, and cursor uint64. Therefore I would have to do an implicit conversion.

aesedepece:
Is `seek` intentionally not using `notOutOfBounds`?

girazoki:
My impression is that with the _relative variable check that needs to be done before checking the length, it makes no sense to put is as a modifier. Besides seek already has a require checking that we do not access out of bounds

Issue #99, opened 2020-05-08
Integer Overflow in ActiveBridgeSetLib.sol


clbartoli:
**Description**
Discovered vulnerability in ActiveBridgeSetlib.sol.
Contingent upon the identityCount of the address received by the _address parameter does not
exist; an overflow will occur setting the identity value to the maximum value possible for UINT16
(65535).

![image](https://user-images.githubusercontent.com/52961116/81419510-a3372c00-914e-11ea-9b7b-757807adbb55.png)

PR #100, opened 2020-05-08
Avoid overflow

### LINKED ISSUE #98: Integer Overflow In BufferLib
**Description**
Discovered vulnerability in BufferLib.sol.
Given the` _offset + _buffer.cursor > MAX (unit256)`, an overflow will occur.

image

clbartoli:
This PR avoids overflows to occur in:

 - BufferLib.sol
 - ActiveBridgeSetLib.sol

Closes #98 and #99

### DIFF
--- a/contracts/ActiveBridgeSetLib.sol
+++ b/contracts/ActiveBridgeSetLib.sol
@@ -164,6 +164,7 @@ library ActiveBridgeSetLib {
   /// @param _abs The Active Bridge Set to be flushed.
   /// @param _address The address to be flushed.
   function flushIdentity(ActiveBridgeSet storage _abs, address _address) private {
+    require(absMembership(_abs, _address) == true, "The address already out of the ARS");
     // Decrement the count of an identity, and if it reaches 0, delete it and update `nextActiveIdentities`count
     _abs.identityCount[_address]--;
     if (_abs.identityCount[_address] == 0) {
--- a/contracts/ActiveBridgeSetLib.sol
+++ b/contracts/ActiveBridgeSetLib.sol
@@ -164,7 +164,7 @@ library ActiveBridgeSetLib {
   /// @param _abs The Active Bridge Set to be flushed.
   /// @param _address The address to be flushed.
   function flushIdentity(ActiveBridgeSet storage _abs, address _address) private {
-    require(absMembership(_abs, _address) == true, "The address already out of the ARS");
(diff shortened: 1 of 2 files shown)


contracts/BufferLib.sol
```diff
@@ -52,6 +52,7 @@ library BufferLib {
     uint64 newCursor = _offset;
     // Deal with relative offsets
     if (_relative == true) {
+      require(newCursor + _buffer.cursor > newCursor && newCursor + _buffer.cursor > _buffer.cursor, "The sum newCursor caused an overflow");
```
girazoki, line 55: I think checking only against one of them is sufficient. Thus just leave the first check in this require

contracts/ActiveBridgeSetLib.sol
```diff
@@ -163,6 +163,7 @@ library ActiveBridgeSetLib {
   /// @param _abs The Active Bridge Set to be flushed.
   /// @param _address The address to be flushed.
   function flushIdentity(ActiveBridgeSet storage _abs, address _address) private {
+    require(absMembership(_abs, _address) == true, "The address already out of the ARS");
```
mariocao, line 166:
```suggestion
    require(absMembership(_abs, _address), "The address already out of the ARS");
```

contracts/BufferLib.sol
```diff
@@ -52,6 +52,7 @@ library BufferLib {
     uint64 newCursor = _offset;
     // Deal with relative offsets
     if (_relative == true) {
+      require(newCursor + _buffer.cursor > newCursor, "The sum newCursor caused an overflow");
```
mariocao, line 55: Maybe a more detailed error message (i.e. stating the function where it occurred)
```suggestion
      require(newCursor + _buffer.cursor > newCursor, "Integer overflow when seeking");
```

contracts/BufferLib.sol
```diff
@@ -68,6 +68,7 @@ library BufferLib {
   function seek(Buffer memory _buffer, uint32 _offset, bool _relative) internal pure returns (uint32) {
     // Deal with relative offsets
     if (_relative) {
+      require(_offset + _buffer.cursor > _offset, "Integer overflow when seeking");
```
aesedepece, line 71: Just for the record: this may be redundant after adding the `notOutOfBounds` modifier from #97

aesedepece: Sorry, I double-checked and it turned out `seek` will not be using `notOutOfBounds`.

PR #102, opened 2020-05-08
Claim verification

### LINKED ISSUE #101: Minor logic error: dr claim verification
**Description**
In reportDataRequestInclusion() it is not verified that the claim of the request has been previously
made. In case the claim is not made, it could burn tokens since pkClaim (the address that will
receive the fee) will be equal to zero. It is recommended to verify that the claim has been made
before calling reportDataRequestInclusion.

clbartoli:
This PR adds a modifier in `ReportDataRequestInclusion` so that reverts if the data request has not been claimed. It also adds a test.

Closes #101

### DIFF
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -97,6 +97,13 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     require(requests[_id].drHash != 0, "DR not yet included");
     _;
   }
+
+  // Ensures the DR inclusion has been already reported
+  modifier drClaimed(uint256 _id) {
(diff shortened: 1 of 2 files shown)


contracts/WitnetRequestsBoard.sol
```diff
@@ -218,6 +219,8 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     external
     drNotIncluded(_id)
  {
+    // Check the data request has been claimed
+    require(dataRequestCanBeClaimed(requests[_id]) == false, "DR not yet claimed");
```
mariocao, line 223:
```suggestion
    require(dataRequestCanBeClaimed(requests[_id]) == false, "Data Request has not yet been claimed");
```

PR #107, opened 2020-05-11
Gas optimizations

aesedepece:
Multiple gas optimizations as suggested by red4sec.

Only three of their suggestions are NOT included in this PR. Explanations below:

- **Witnet-ethereum-bridge\contracts\Witnet.sol:432-433 - Could be uint16 instead of uint64.**
        It would be antiergonomic to force the caller to do the uint64 → uint16 trimming, because this is always used on uint64 values because that's the API of our CBOR arrays.

- **Witnet-ethereum-bridge\contracts\Witnet.sol:299-300 - Use bytes instead of uint64.**
        It is a convention in Witnet that errors are returned as CBOR arrays containing unsigned integers. As the only unsigned integer type currently supported is `uint64`, we are force to use that.
        If you want the raw bytes representation of the error, you already have `result.cborValue.buffer.data`

- **Witnet-ethereum-bridge\contracts\WitnetRequestsBoard.sol:149** - Delete requests once their result is read.
        Just not yet (requires discussion and many changes on multiple repos and docs)

### DIFF
--- a/contracts/BufferLib.sol
+++ b/contracts/BufferLib.sol
@@ -9,7 +9,7 @@ pragma solidity >=0.5.3 <0.7.0;
 library BufferLib {
   struct Buffer {
     bytes data;
-    uint64 cursor;
+    uint32 cursor;
   }
 
   /**
@@ -18,12 +18,12 @@ library BufferLib {
   * @param _length How many bytes to read and consume from the buffer.
   * @return A `bytes memory` containing the first `_length` bytes from the buffer, counting from the cursor position.
   */
-  function read(Buffer memory _buffer, uint64 _length) internal pure returns (bytes memory) {
+  function read(Buffer memory _buffer, uint32 _length) internal pure returns (bytes memory) {
     // Make sure not to read out of the bounds of the original bytes
     require(_buffer.cursor + _length <= _buffer.data.length, "Not enough bytes in buffer when reading");
     // Create a new `bytes memory` value and copy the desired bytes into it
     bytes memory value = new bytes(_length);
-    for (uint64 index = 0; index < _length; index++) {
+    for (uint32 index = 0; index < _length; index++) {
       value[index] = next(_buffer);
     }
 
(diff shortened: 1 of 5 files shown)


contracts/BufferLib.sol
```diff
@@ -18,16 +18,30 @@ library BufferLib {
   * @param _length How many bytes to read and consume from the buffer.
   * @return A `bytes memory` containing the first `_length` bytes from the buffer, counting from the cursor position.
   */
-  function read(Buffer memory _buffer, uint64 _length) internal pure returns (bytes memory) {
+  function read(Buffer memory _buffer, uint32 _length) internal pure returns (bytes memory) {
```
aesedepece, line 21: BREAKING CHANGE

contracts/BufferLib.sol
```diff
@@ -48,15 +62,15 @@ library BufferLib {
   * buffer (`true`).
   * @return The final position of the cursor (will equal `_offset` if `_relative` is `false`).
   */
-  function seek(Buffer memory _buffer, uint64 _offset, bool _relative) internal pure returns (uint64) {
-    uint64 newCursor = _offset;
+  // solium-disable-next-line security/no-assign-params
+  function seek(Buffer memory _buffer, uint32 _offset, bool _relative) internal pure returns (uint32) {
```
aesedepece, line 66: BREAKING CHANGE

contracts/BufferLib.sol
```diff
@@ -67,7 +81,7 @@ library BufferLib {
   * @param _relativeOffset How many bytes to move the cursor forward.
   * @return The final position of the cursor.
   */
-  function seek(Buffer memory _buffer, uint64 _relativeOffset) internal pure returns (uint64) {
+  function seek(Buffer memory _buffer, uint32 _relativeOffset) internal pure returns (uint32) {
```
aesedepece, line 84: BREAKING CHANGE

contracts/WitnetRequestsBoard.sol
```diff
@@ -140,29 +140,36 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
...
+
+    // Create a new `DataRequest` object and initialize all the fields
+    DataRequest memory request;
+    request.dr = _serialized;
+    request.inclusionReward = SafeMath.sub(msg.value, _tallyReward);
+    request.tallyReward = _tallyReward;
+    request.result = "";
+    request.timestamp = 0;
+    request.drHash = 0;
+    request.pkhClaim = address(0);
```
girazoki, line 164: shall we remove this? I think it should already initialize to those values

contracts/BufferLib.sol
mariocao, line 21: Maybe an explanation in the why an `uint32` is enough. It supports to read up to 537MB which is already...a lot :smile:

aesedepece:
Suggestions are incorporated and branch is rebased.

PR #113, opened 2020-05-21
fix: update ABS activity checks

girazoki:
This issue moves checking whether the block number provided is higher than the last one inserted to the ABS library.Closes issue 13.10 from the security audit

### DIFF
--- a/contracts/ActiveBridgeSetLib.sol
+++ b/contracts/ActiveBridgeSetLib.sol
@@ -28,8 +28,8 @@ library ActiveBridgeSetLib {
     uint256 lastBlockNumber;
   }
 
-  modifier validBlockNumber(uint256 _currentBlock, uint256 _lastBlock) {
-    require (_currentBlock >= _lastBlock, "The last updated block was higher than the one provided");
+  modifier validBlockNumber(uint256 _blockFromArguments, uint256 _blockFromContractState) {
+    require (_blockFromArguments >= _blockFromContractState, "The provided block is older than the last updated block");
     _;
   }
 
--- a/test/TestActiveBridgeSet.sol
+++ b/test/TestActiveBridgeSet.sol
@@ -138,7 +138,7 @@ contract TestActiveBridgeSet {
     fakeAbs.lastBlockNumber = 1;
     TestActiveBridgeSet(address(throwProxy)).errorUpdateABS(0);
     (r, error) = throwProxy.execute.gas(100000)();
-    Assert.isFalse(r, "The last updated block was higher than the one provided");
+    Assert.isFalse(r, "The provided block is older than the last updated block");
   }
 
   function testNotValidPushBlockNumber() external {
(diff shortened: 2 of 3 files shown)


contracts/ActiveBridgeSetLib.sol
```diff
@@ -28,10 +28,18 @@ library ActiveBridgeSetLib {
     uint256 lastBlockNumber;
   }
 
+  modifier validBlockNumber(uint256 _currentBlock, uint256 _lastBlock) {
+    require (_currentBlock >= _lastBlock, "The last block number updated was higher than the one provided");
```
mariocao, line 32:
```suggestion
    require (_currentBlock >= _lastBlock, "The last updated block was higher than the one provided");
```

contracts/WitnetRequestsBoard.sol
```diff
@@ -347,7 +347,6 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
   /// @dev Updates the ABS activity with the block number provided.
   /// @param _blockNumber update the ABS until this block number.
   function updateAbsActivity(uint256 _blockNumber) external {
-    require (_blockNumber >= abs.lastBlockNumber, "The last block number updated was higher than the one provided");
     require (_blockNumber <= block.number, "The block number provided has not been reached");
```
mariocao, line 350:
```suggestion
    require (_blockNumber <= block.number, "The provided block number has not been reached");
```

test/TestActiveBridgeSet.sol
```diff
@@ -21,6 +21,7 @@ contract TestActiveBridgeSet {
   ];
 
   ActiveBridgeSetLib.ActiveBridgeSet internal abs;
+  ActiveBridgeSetLib.ActiveBridgeSet internal fakeabs;
```
mariocao, line 24:
```suggestion
  ActiveBridgeSetLib.ActiveBridgeSet internal fakeAbs;
```

contracts/ActiveBridgeSetLib.sol
```diff
@@ -28,10 +28,18 @@ library ActiveBridgeSetLib {
     uint256 lastBlockNumber;
   }
 
+  modifier validBlockNumber(uint256 _currentBlock, uint256 _lastBlock) {
+    require (_currentBlock >= _lastBlock, "The last updated block was higher than the one provided");
```
aesedepece, line 32: Do you mind if we go the other way so the message explicitly complains about the bad argument?
```suggestion
    require (_currentBlock >= _lastBlock, "The provided block is older than the last updated block");
```

contracts/ActiveBridgeSetLib.sol
```diff
@@ -28,10 +28,18 @@ library ActiveBridgeSetLib {
     uint256 lastBlockNumber;
   }
 
+  modifier validBlockNumber(uint256 _currentBlock, uint256 _lastBlock) {
```
aesedepece, line 31: `current` and `last` are confusing to me in this context, as they could be seen as "currently in the contract state" or "last appended to the chain" (and both interpretations are wrong).

What about this?
```suggestion
  modifier validBlockNumber(uint256 _blockFromArguments, uint256 _blockFromContractState) {
```

Then the `require` instruction below would become super expressive as it would literally read _"require that block from arguments is greater or equal than block from contract state"_.

PR #133, opened 2020-12-01
WRB: store request address instead of request bytecode

### LINKED ISSUE #131: Avoid storing data request bytecode in WRB
The storage of data request bytecode is not needed to be located in the WRB.
Instead the WRB should retrieve it from the user Contract (`is UsingWitnet`).
This should decrease some gas.

clbartoli:
The DataRequest structure in the contract WRB stores the request address sent by UsingWitnet. With this change the WRB does not longer need to store the request bytecode.

Closes #131


contracts/WitnetRequestsBoard.sol
```diff
@@ -164,7 +165,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
 
     // Create a new `DataRequest` object and initialize all the non-default fields
     DataRequest memory request;
-    request.dr = _serialized;
+    request.requestAddress = _requestAddress;
```
girazoki, line 168: We should store the hash of the DR to avoid potential attacks, as     uint256 drOutputHash = uint256(sha256(requestContract.bytecode())). Then we should use this value when the PoI is demonstrated

contracts/WitnetRequestsBoard.sol
```diff
@@ -316,7 +317,8 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
   /// @return The result of the data request as bytes.
   function readDataRequest(uint256 _id) external view returns(bytes memory) {
     require(requests.length > _id, "Id not found");
-    return requests[_id].dr;
+    Request requestContract = Request(requests[_id].requestAddress);
```
girazoki, line 320: Potentially we can add a check here chekcing that the hash coincides

girazoki (approved):

PR #141, opened 2020-12-16
gas: add 9 poi levels to gas analysis and activity removal

girazoki:
.This PR pushes the limits of the gas analysis, considering 9 PoI levels for both result and DR inclusion, plus activity cleaning for every slot. I also took the chance of reducing the activity slots from 100 to 50, as I think it might be more efficient for the numbers we might manage at this point. In the WRB, I wrote than all these findings as maximum gas requirements for each of the actions a bridge node needs to do


contracts/WitnetRequestsBoard.sol
```diff
@@ -25,6 +25,18 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
   // This should be at least superblock_period*2*checkpoint_period/ethereum_block_time
   // This yields 60, we double it to be conservative
   uint256 public constant CLAIM_EXPIRATION = 120;
+  
+  // Max gas values as calculate with gas-analysis
+  // Claiming is not subject to substantial increases as it is only composed of a VRF verification 
+  uint256 public constant MAX_CLAIM_DR_GAS = 215248;
```
mariocao, line 31: Is the gas cost table is not updated? :P

test/wrb_gas.js
```diff
@@ -292,6 +313,74 @@ contract("WitnetRequestBoard", ([
     })
   })
 
+  describe("report data request inclusion pushing the limtis (9 PoI levels and activity to be removed)", async () => {
```
mariocao, line 316: Niceeee! :smile:

mariocao (approved):

PR #143, opened 2020-12-18
Pay block reward

girazoki:
This PR expands on https://github.com/witnet/witnet-ethereum-block-relay/pull/29 to be able to pay the block relayer. The block reward is divided in two, one for the inclusion and one for the reward.

 Missing things:

- Upgrade Request should also include an upgrade for the block relay reward.
- More extensive testing

### DIFF
--- a/contracts/UsingWitnet.sol
+++ b/contracts/UsingWitnet.sol
@@ -33,9 +33,10 @@ contract UsingWitnet {
   }
 
   // Ensures that user-specified rewards are equal to the total transaction value to prevent users from burning any excess value
-  modifier validRewards(uint256 _requestReward, uint256 _resultReward) {
-    require(_requestReward + _resultReward >= _requestReward, "The sum of rewards overflows");
-    require(msg.value == _requestReward + _resultReward, "Transaction value should equal the sum of rewards");
+  modifier validRewards(uint256 _requestReward, uint256 _resultReward, uint256 _blockReward) {
+    require(_requestReward + _resultReward + _blockReward >= _requestReward, "The sum of rewards overflows");
+
+    require(msg.value == _requestReward + _resultReward + _blockReward, "Transaction value should equal the sum of rewards");
     _;
   }
 
@@ -47,12 +48,12 @@ contract UsingWitnet {
   * @param _resultReward Reward specified for the user which posts back the request result
   * @return Sequencial identifier for the request included in the WitnetRequestsBoard
   */
-  function witnetPostRequest(Request _request, uint256 _requestReward, uint256 _resultReward)
+  function witnetPostRequest(Request _request, uint256 _requestReward, uint256 _resultReward, uint256 _blockReward)
     internal
-    validRewards(_requestReward, _resultReward)
+    validRewards(_requestReward, _resultReward, _blockReward)
   returns (uint256)
   {
-    return wrb.postDataRequest{value: _requestReward + _resultReward}(address(_request), _resultReward);
+    return wrb.postDataRequest{value: _requestReward + _resultReward + _blockReward}(address(_request), _requestReward, _resultReward);
   }
 
   /**
@@ -79,7 +80,7 @@ contract UsingWitnet {
   */
   function witnetUpgradeRequest(uint256 _id, uint256 _requestReward, uint256 _resultReward)
     internal
-    validRewards(_requestReward, _resultReward)
+    validRewards(_requestReward, _resultReward, 0)
   {
     wrb.upgradeDataRequest{value: msg.value}(_id, _resultReward);
   }
--- a/contracts/UsingWitnet.sol
+++ b/contracts/UsingWitnet.sol
@@ -34,8 +34,9 @@ contract UsingWitnet {
 
   // Ensures that user-specified rewards are equal to the total transaction value to prevent users from burning any excess value
   modifier validRewards(uint256 _requestReward, uint256 _resultReward, uint256 _blockReward) {
-    require(_requestReward + _resultReward + _blockReward >= _requestReward, "The sum of rewards overflows");
-
+    uint256 reqResReward = _requestReward + _resultReward;
+    require(reqResReward >= _requestReward, "The sum of rewards overflows");
+    require(reqResReward + _blockReward >= reqResReward, "The sum of rewards overflows");
     require(msg.value == _requestReward + _resultReward + _blockReward, "Transaction value should equal the sum of rewards");
     _;
   }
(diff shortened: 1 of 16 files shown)


contracts/UsingWitnet.sol
```diff
@@ -33,9 +33,10 @@ contract UsingWitnet {
   }
 
   // Ensures that user-specified rewards are equal to the total transaction value to prevent users from burning any excess value
-  modifier validRewards(uint256 _requestReward, uint256 _resultReward) {
-    require(_requestReward + _resultReward >= _requestReward, "The sum of rewards overflows");
-    require(msg.value == _requestReward + _resultReward, "Transaction value should equal the sum of rewards");
+  modifier validRewards(uint256 _requestReward, uint256 _resultReward, uint256 _blockReward) {
+    require(_requestReward + _resultReward + _blockReward >= _requestReward, "The sum of rewards overflows");
```
tmpolaczyk, line 37: This overflow check is wrong, with the values `_requestReward = 1`, `_resultReward = _blockReward = 2**256 - 1` it returns true:

    require(1 + (2**256 - 1) + (2**256 - 1) >= 1)

So use a different overflow check, and add tests.

clbartoli:
If I got it right, the `blockReward` is defined when posting the data request as the sum of the `inclusionReward` and the `tallyReward`. If this is correct, then there is no need to include the `blockReward` as input in any function, since it can be calculated from the other rewards.

Having this in mind, some modifiers change
- `payingEnough` in WitnetRequestBoard.sol should make sure the value is greater or equal `2*_tally`, the reward of the tally and  the block reward.
- `ValidRewards` in UsingWitnet.sol has to consider double the rewards, something like 
  `require(_requestReward + _resultReward >= _requestReward && 2*(_requestReward +_requestReward) >= 2*_requestReward, "The sum of rewards overflows");

The tests will need to be changed as well, as some of the values of the transactions are not longer valid.`

contracts/UsingWitnet.sol
```diff
@@ -47,12 +48,12 @@ contract UsingWitnet {
   * @param _resultReward Reward specified for the user which posts back the request result
   * @return Sequencial identifier for the request included in the WitnetRequestsBoard
   */
-  function witnetPostRequest(Request _request, uint256 _requestReward, uint256 _resultReward)
+  function witnetPostRequest(Request _request, uint256 _requestReward, uint256 _resultReward, uint256 _blockReward)
```
mariocao, line 51: The `@param` documentation is missing :)

contracts/WitnetRequestsBoard.sol
```diff
@@ -166,10 +167,10 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
   /// @param _requestAddress The request contract address which includes the request bytecode.
   /// @param _tallyReward The amount of value that will be detracted from the transaction value and reserved for rewarding the reporting of the final result (aka tally) of the data request.
   /// @return The unique identifier of the data request.
-  function postDataRequest(address _requestAddress, uint256 _tallyReward)
+  function postDataRequest(address _requestAddress, uint256 _inclusionReward, uint256 _tallyReward)
```
mariocao, line 170: Missing `@param` documentation

contracts/WitnetRequestsBoard.sol
```diff
@@ -198,7 +200,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
   /// @dev Increments the rewards of a data request by adding more value to it. The new request reward will be increased by msg.value minus the difference between the former tally reward and the new tally reward.
   /// @param _id The unique identifier of the data request.
   /// @param _tallyReward The new tally reward. Needs to be equal or greater than the former tally reward.
-  function upgradeDataRequest(uint256 _id, uint256 _tallyReward)
+  function upgradeDataRequest(uint256 _id,  uint256 _tallyReward)
```
mariocao, line 203:
```suggestion
  function upgradeDataRequest(uint256 _id, uint256 _tallyReward)
```

contracts/WitnetRequestsBoard.sol
```diff
@@ -274,6 +276,10 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
       _epoch,
       _index,
       requests[_id].drOutputHash), "Invalid PoI");
+    address payable relayer = payable(blockRelay.readRelayerAddress(_blockHash, _epoch));
+    uint256 relayerPayment = requests[_id].blockReward/2;
```
mariocao, line 280: Are relay DR inclusion and block similar in terms of gas cost?
Just asking myself if the factor 2 makes sense or could be improved...

mariocao, line 280:
```suggestion
    uint256 relayerPayment = requests[_id].blockReward / 2;
```

contracts/WitnetRequestsBoard.sol
```diff
@@ -274,6 +276,10 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
...
+    requests[_id].blockReward = requests[_id].blockReward - relayerPayment; 
+    relayer.transfer(relayerPayment);
```
mariocao, line 282:
```suggestion
    address payable relayer = payable(blockRelay.readRelayerAddress(_blockHash, _epoch));
    uint256 relayerPayment = requests[_id].blockReward/2;
    requests[_id].blockReward = requests[_id].blockReward - relayerPayment; 
    relayer.transfer(relayerPayment);
```
Add comment explaining that we are storing reward and doing transfer :)

contracts/mocks/MockWitnetRequestsBoard.sol
```diff
@@ -39,7 +40,7 @@ contract MockWitnetRequestsBoard is WitnetRequestsBoardInterface {
   /// @param _requestAddress The request contract address which includes the request bytecode.
   /// @param _tallyReward The amount of value that will be detracted from the transaction value and reserved for rewarding the reporting of the final result (aka tally) of the data request.
   /// @return The unique identifier of the data request.
-  function postDataRequest(address _requestAddress, uint256 _tallyReward)
+  function postDataRequest(address _requestAddress, uint256 _inclusionReward, uint256 _tallyReward)
```
mariocao, line 43: Missing `@param` doc

contracts/WitnetRequestsBoard.sol
girazoki: In theory they are both block inclusions (one for DR and the other one for result), so they should be yes.

mariocao (approved):

PR #145, opened 2021-01-14
Store gas price per dr

girazoki:
This PR:

- Stores the gas price set by the requestor in the DR information structure.
- Adds a getter to retrieve such a gas price.
- Rejects the DR if the rewards do not match the expected gas times the gas price set.

### DIFF
--- a/contracts/WitnetRequestsBoard.sol
+++ b/contracts/WitnetRequestsBoard.sol
@@ -389,7 +389,7 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
 
   /// @dev Retrieves the gas price set for a specific DR ID.
   /// @param _id The unique identifier of the data request.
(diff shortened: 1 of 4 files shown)


contracts/WitnetRequestsBoard.sol
```diff
@@ -368,6 +387,13 @@ contract WitnetRequestsBoard is WitnetRequestsBoardInterface {
     return requests[_id].result;
   }
 
+  /// @dev Retrieves the gas price set for a specific DR ID.
+  /// @param _id The unique identifier of the data request.
+  /// @return The gas price set by the request creator
```
clbartoli, line 392:
```suggestion
  /// @return The gas price set by the request creator.
```

mariocao (approved):

clbartoli (approved):

PR #153, opened 2021-02-15
chore: update solhint and add apikey to truffle config

mariocao:


truffle-config.js
```diff
@@ -27,8 +27,9 @@ module.exports = {
   mocha: {
     reporter: 'eth-gas-reporter',
     reporterOptions: {
+        coinmarketcap: process.env.COINMARKETCAP_API_KEY,
         currency: "USD",
-        gasPrice: 50,
+        gasPrice: 100,
```
mariocao, line 32: This `gasPrice` is set to 100 just to be the same as the one used in the `witnet-ethereum-block-relay` repository

clbartoli (approved):

PR #163, opened 2021-05-10
feat: add message for bridge radon errors

clbartoli:
Add error messages for bridgeErrors

### DIFF
--- a/contracts/Witnet.sol
+++ b/contracts/Witnet.sol
@@ -402,7 +402,8 @@ library Witnet {
    * @param _result An instance of `Result`.
    * @return A tuple containing the `CBORValue.Error memory` decoded from the `Result`, plus a loggable error message.
    */
-  function asErrorMessage(Result memory _result) external pure returns(ErrorCodes, string memory) {
+
+  function asErrorMessage(Result memory _result) public pure returns (ErrorCodes, string memory) {
     uint64[] memory error = asRawError(_result);
     if (error.length == 0) {
       return (ErrorCodes.Unknown, "Unknown error (no error code)");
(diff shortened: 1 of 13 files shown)


contracts/Witnet.sol
```diff
@@ -491,7 +491,14 @@ library Witnet {
...
         " stage"
       );
-    } else {
+      } else if (errorCode == ErrorCodes.BridgeMalformedRequest) {
+      errorMessage = "The structure of the request is invalid and it can not be parsed";
+      } else if (errorCode == ErrorCodes.BridgePoorIncentives) {
+      errorMessage = "The wit cost of the request exceeds the limit";
+      } else if (errorCode == ErrorCodes.BridgeOversizedResult) {
+      errorMessage = "The request result length exceeds the limit";
+      }
```
mariocao, line 500:
```suggestion
      } else if (errorCode == ErrorCodes.BridgeMalformedRequest) {
          errorMessage = "The structure of the request is invalid and it cannot be parsed";
      } else if (errorCode == ErrorCodes.BridgePoorIncentives) {
          errorMessage = "The wit cost of the request exceeds the limit";
      } else if (errorCode == ErrorCodes.BridgeOversizedResult) {
          errorMessage = "The request result length exceeds a bridge contract defined limit";
      }
```

Indentation :)

I don't know, the second error message poor incentives is not completely accurate because it could failed because of  the collateral. The documentation in `witnet-rust` for that error was:

```
/// The request is rejected on the grounds that it may cause the submitter to spend or stake an
/// amount of value that is unjustifiably high when compared with the reward they will be getting
```

mariocao (approved):

contracts/Witnet.sol
```diff
@@ -411,88 +411,94 @@ library Witnet {
...
+          " from ",
+          stageName(error[1]),
+          " stage"
+        );
+    } else if (errorCode == ErrorCodes.BridgeMalformedRequest) {
+        errorMessage = "The structure of the request is invalid and it cannot be parsed";
+    } else if (errorCode == ErrorCodes.BridgePoorIncentives) {
+        errorMessage = "The request has been rejected by the bridge node due to poor incentives";
+    } else if (errorCode == ErrorCodes.BridgeOversizedResult) {
+        errorMessage = "The request result length exceeds a bridge contract defined limit";
```
aesedepece, line 499: In this one, we could even interpolate the actual and limit lengths.

aesedepece (approved):

PR #168, opened 2021-06-22
Chore: reorganize solidity files into subfolders

guidiaz:
Convenient reorganization of source files as to approach abstractions and library decouplings and inheritance-based implementations.


clbartoli (approved):

.gitignore
```diff
@@ -7,6 +7,8 @@ contracts/flattened
 .env
 .idea
 .vscode
+.old
+*.old
```
mariocao, line 11: Isn't it `.old` and `*.old` the same in this context?

tmpolaczyk: `*.old` also includes `a.old`, `b.old`, etc.

tmpolaczyk: I personally prefer to keep the gitignore as clean as possible, because you can set up a [global gitignore](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) if you need to exclude certain files created by your editor.

tmpolaczyk (approved):

.gitignore
mariocao: Yep, I agree. Let's delete them from the gitignore :P

guidiaz:
Superseded by https://github.com/witnet/witnet-ethereum-bridge/pull/176

PR #172, opened 2021-06-23
Refactor: follow delegate-proxy approach, but keeping `WitnetRequestBoardInterface` as-is and avoiding breaking changes on the bridge node

guidiaz:
Keeping `WitnetRequestBoardInterface` as-is, with no-extra interface declarations and avoiding breaking changes on the bridge node, by now.

Delegate-proxy approach enables:

- Future implementation of legitimate DR destruction (from DR requestor only).
- Future upgrades to data model.
- Future provisioning of up-to-date result decoder from WRB instance.
- Eventual destruction of WRB instances.

Depends on PR #170.


contracts/WitnetProxy.sol
```diff
@@ -0,0 +1,45 @@
...
+  constructor () {}
+
+  receive() external payable {
+    revert("WitnetProxy: no ETH accepted");
+  }
+
+  // solhint-disable-next-line
+  fallback() external payable {
+    address _wrb = address(wrb);
+    assembly {
```
tmpolaczyk, line 19: :warning: assembly :warning: 

I would appreciate a comment explaining why is this assembly block needed and why it is safe.

contracts/data/WitnetBoardData.sol
```diff
@@ -2,9 +2,55 @@
...
+    assembly {
+      _struct.slot := WITNET_BOARD_DATA_SLOTHASH
+    }
+  }
+
+  function __dataRequest(uint256 id) internal view returns (WitnetTypes.DataRequest storage) {
+    return __data().requests[id].dr;
+  }
+  
+  bytes32 internal constant WITNET_BOARD_DATA_SLOTHASH = "WitnetBoard.Data";
```
tmpolaczyk, line 54: I don't understand what is this, it's called "hash" but it is a plaintext string?

contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -0,0 +1,278 @@
...
+    uint256 internal constant __ESTIMATED_REPORT_RESULT_GAS = 102496;
+
+    constructor() {
+        // sets instance as initialized:
+        __data().instance = address(this);
+    }
+
+
+    // ================================================================================================================
+    // --- Overrides 'Destructible' -----------------------------------------------------------------------------------
```
tmpolaczyk, line 38: I don't like this pattern of separating implementations using comments because it's very fragile. I prefer interfaces over inheritance, but not sure if it makes sense here.

contracts/data/WitnetBoardData.sol
```diff
@@ -0,0 +1,56 @@
...
+    require(__dataRequest(id).txhash == 0, "WitnetBoardData: already solved");
+    _;
+  }
+
+  modifier wasPosted(uint256 id) {
+    require(id > 0 && id <= __data().noRequests, "WitnetBoardData: not yet posted");
+    _;
+  }
+
+  function __data() internal pure returns (SWitnetBoardData storage _struct) {
```
mariocao, line 44:
```suggestion
  function _data() internal pure returns (SWitnetBoardData storage _struct) {
```

Afaik, the widely used convention is only 1 underscore for internal functions.
Still, it's nothing official after reviewing the [style guide](https://docs.soliditylang.org/en/latest/style-guide.html) from the solidity documentation.

For example, in the `ERC20` there is an underscore for internal functions, e.g. `_approve`, `_transfer`, ...

contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -0,0 +1,278 @@
...
+    // --- Utility functions not declared within an interface ---------------------------------------------------------
+
+    function setReporters(address[] memory reporters)
+        public
+        virtual
+        onlyOwner
+    {
+        for (uint ix = 0; ix < reporters.length; ix ++) {
+            address _reporter = reporters[ix];
+            if (_reporter != address(0)) {
```
tmpolaczyk, line 86: Why this check?

contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -0,0 +1,278 @@
...
+    function version() virtual external pure override returns (string memory) {
+        return "WitnetRequestBoard-Centralized-v0.3.0";
+    }
+
+
+    // ================================================================================================================
+    // --- Utility functions not declared within an interface ---------------------------------------------------------
+
+    function setReporters(address[] memory reporters)
+        public
```
tmpolaczyk, line 80: This function is public and it's missing documentation

contracts/data/WitnetBoardData.sol
```diff
@@ -2,9 +2,55 @@
...
  * @author Witnet Foundation
  */
-abstract contract WitnetBoardData {}
+abstract contract WitnetBoardData {  
+
+  struct SWitnetBoardData {
+    address owner;
+    address instance;
+    mapping (uint => SWitnetBoardDataRequest) requests;
+    uint256 noRequests;
```
tmpolaczyk, line 17: `noRequests` is a really bad name to indicate number of requests. Try `numRequests` or `requestsCount`.

contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -0,0 +1,278 @@
...
+        virtual
+        wasPosted(id)
+        returns (bytes memory _bytecode)
+    {
+        WitnetTypes.DataRequest storage _dr = __dataRequest(id);
+        _bytecode = WitnetRequest(_dr.addr).bytecode();
+        require(
+            WitnetTypes.computeDataRequestCodehash(_bytecode) == _dr.codehash,
+            "WitnetRequestBoard: manipulated request"
+        );
```
tmpolaczyk, line 118: Could this function be implemented as

```js
return readDr(id).bytecode();
```

or similar? Or it is more expensive?

contracts/libs/BufferLib.sol
```diff
@@ -2,7 +2,7 @@
 
 pragma solidity >=0.7.0 <0.9.0;
 
-
+import "../exports/WitnetTypes.sol";
```
tmpolaczyk, line 5: Moving the `Buffer` struct to `WitnetTypes` makes this library depend on `WitnetTypes`. I think that this should be a standalone library, with the `Buffer` struct defined here.

contracts/libs/CBOR.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity >=0.7.0 <0.9.0;
 pragma experimental ABIEncoderV2;
 
 import "./BufferLib.sol";
-
+import "../exports/WitnetTypes.sol";
```
tmpolaczyk, line 7: Same as the `Buffer` comment, `Value` is a type that belongs to the `CBOR` library. It doesn't make sense to import it from somewhere else.

contracts/WitnetProxy.sol
```diff
@@ -0,0 +1,45 @@
...
+      let result := delegatecall(gas(), _wrb, ptr, calldatasize(), 0, 0)
+      let size := returndatasize()
+      returndatacopy(ptr, 0, size)
+      switch result
+        case 0  { revert(ptr, size) }
+        default { return(ptr, size) }
+    }
+  }
+
+  function upgrade(address _newWrb, bytes memory _initData) public {
```
tmpolaczyk, line 32: I can't find the `onlyOwner` check. Does that mean that anyone can upgrade the `WitnetProxy`?

test/helpers/WitnetRequestBoardTestHelper.sol
```diff
@@ -40,66 +25,43 @@ contract WitnetRequestBoardTestHelper is IWitnetRequestor {
     upgradable = _upgradable;
 
     // Insert an empty request so as to initialize the requests array with length > 0
-    DataRequest memory request;
-    requests.push(request);
-    committee = _committee;
+    setReporters(_committee);
```
tmpolaczyk, line 28: This comment is now outdated

test/using_witnet.js
```diff
@@ -1,7 +1,13 @@
+const {
+  // expectEvent,
+  expectRevert,
+} = require("@openzeppelin/test-helpers")
+// const { expect } = require("chai")
```
tmpolaczyk, line 5: Remove comments?

test/wrb.js
```diff
@@ -299,7 +297,7 @@ contract("WitnetRequestBoard", ([
           requestId, 0, resultHex,
           { from: owner, gasPrice: 1 }
         ),
-        "Data request transaction cannot be zero"
+        "tx hash cannot be zero"
```
tmpolaczyk, line 300: The old message was better because `tx hash` can also mean ethereum transaction hash, or tally transaction hash. If the string length is a problem, try "DR tx hash cannot be zero".

test/wrb.js
```diff
@@ -360,13 +358,13 @@ contract("WitnetRequestBoard", ([
           from: other,
           gasPrice: 1,
         }),
-        "The dr has been manipulated and the bytecode has changed"
+        "manipulated request"
```
tmpolaczyk, line 361: This was confusing to me: "manipulated" means that the data request bytecode was manipulated, not that the result was manipulated. I would prefer a message like "data request bytecode has changed", because "manipulated" gives the impression that witnet can be manipulated easily and it's not secure.

test/wrb.js
```diff
@@ -394,14 +392,50 @@ contract("WitnetRequestBoard", ([
   describe("estimate gas cost", async () => {
     it("anyone can estime a data request gas cost", async () => {
       // Gas price = 1
-      const maxResRe = new BN(await this.WitnetRequestBoard.ESTIMATED_REPORT_RESULT_GAS.call())
+      const maxResRe = new BN(102496)
```
tmpolaczyk, line 395: Why hardcode the number here?

contracts/data/WitnetBoardData.sol
guidiaz: The text "WitnetBoard.Data" is actually converted to `bytes32`, which aftewards is interpreted as a slot storage pointer (ie. storage slot hash). Therefore, maximum name length would be 32.

This approach enables the possibility to let a contract report supported storage slots in a human-readable way, without the caller needing a lookup service (either onchain or offchain) as to translate hashes to actual data struct names. 

Two things to have in mind, though: 
1. Storage slot name maximum length is 32.
2. Actual storage slot size must be smaller than `2 ^ (32 - len(slot_name) - 1` if a storage slot name is to be eventually reused in further data model extensions (e.g. `WitnetBoard.Data2`).

contracts/WitnetProxy.sol
guidiaz: It will be done once merged all related PRs get merged into a new and single branch (e.g. "v0.3")

contracts/impls/centralized/WitnetRequestBoard.sol
guidiaz: Just to make life of eventual code ammenders easier. I agree implementations blocks have to be preserved whenever inherited interfaces/contracts are changed, or new ones added.

contracts/data/WitnetBoardData.sol
guidiaz: `numRequests` will be

contracts/data/WitnetBoardData.sol
guidiaz: I'd rather prefer to used double underscores for internal functions, as to avoid possible confusion when calling libraries on passed arguments (which are usually prefixed with one single underscore). We do agree however that internal functions should be prefixed w/ at least one underscore :-)

contracts/impls/centralized/WitnetRequestBoard.sol
guidiaz: Indeed, it may not be needed in current implementation, as the reporters list is only used for checking ` msg.sender`, which it's true can never be zero. Removing this check on elements of passed array decreases gas cost of calling `setReporters(address[])` with large arrays.

contracts/impls/centralized/WitnetRequestBoard.sol
guidiaz: Actually, `readDr(uint)` returns a struct, not a contract. But yes, you are right if:
1.  `__checkDr(uint)` is changed as to check for bytecode integrity only if the DR was not deleted. 
2. `readDr(uint)` is declared as `public` instead of `external`.

test/wrb.js
guidiaz: Right, it'll change it to: "DR tally tx hash cannot be zero"

test/wrb.js
guidiaz: `WRB.ESTIMATED_REPORT_RESULT_GAS` constant field was actually hardcoded in the same way. It seemed to be have a temporary and very specifc use before a better estimation mechanism is adopted. Having `WRB.estimateGasCost(uint256)` is available if needed.

test/wrb.js
guidiaz: Ok. This "manipulated request" text is part of revert messages produced at the "WitnetRequestBoard" contract. It will be changed to "bytecode changed after posting"

contracts/WitnetProxy.sol
guidiaz: So right. `upgrade` function should delegate call to `Upgradable(delegate).canUpgradeFrom(msg.sender)`, instead of surrogate call to `Upgradable(delegate).isUpgradable()`, to assert whether the contract if upgradable from `msg.sender`, before delegating call to `Upgradable(newDelegate).initialize(bytes)`.

Note: new implementations can reuse stored data from previous implementations, as they all run in the context of the proxy address, so it is up to new implementations to require "ownership" or any other role from `msg.sender` when processing the `initialize(bytes)` method.

guidiaz: New unitary tests checking these situations will be added.

contracts/libs/BufferLib.sol
guidiaz: Sorry, this is the way to go if the Witnet library is to be dicoupled from the "UsingWitnet" abstract contract, and therefore avoding third party developers to be obliged to link to a (possibly outdated) pre-deployed Witnet library.

contracts/libs/CBOR.sol
guidiaz: Same as before.

contracts/UsingWitnet.sol
```diff
@@ -3,26 +3,23 @@
...
-  WitnetRequestBoardInterface internal immutable wrb;
+  WitnetRequestBoardInterface internal immutable witnet;
 
  /**
   * @notice Include an address to specify the WitnetRequestBoard.
   * @param _wrb WitnetRequestBoard address.
   */
   constructor(address _wrb) {
-    wrb = WitnetRequestBoardInterface(_wrb);
+    require(_wrb != address(0), "UsingWitnet: null contract");
```
clbartoli, line 21: "WitnetRequestBoard: null contract"?

clbartoli: could you add a test for this require?

guidiaz:
Supersed by https://github.com/witnet/witnet-ethereum-bridge/pull/176

PR #173, opened 2021-06-24
Feat: enable destruction of solved data requests

guidiaz:
Keeping just one interface definition (i.e. `WitnetRequestBoardInterface`).

Depends on:
- https://github.com/witnet/witnet-ethereum-bridge/pull/172

Contributes to issues:
- https://github.com/witnet/internal/issues/65
- https://github.com/witnet/witnet-ethereum-bridge/issues/167

Solves https://github.com/witnet/internal/issues/76.


contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -205,7 +182,23 @@ contract WitnetRequestBoard
...
+    /// @param id The unique identifier of a previously posted data request.
+    /// @return _result The CBOR-encoded result of the DR.
+    function destroyResult(uint256 id)
+        external
+        virtual override
+        returns (bytes memory _result)
+    {
+        SWitnetBoardDataRequest storage _req = __data().requests[id];
+        require(msg.sender == _req.dr.requestor, "WitnetRequestBoard: only actual requestor");
+        require(_req.dr.txhash != 0, "WitnetReuqestBoard: not yet solved");
```
tmpolaczyk, line 196:
```suggestion
        require(_req.dr.txhash != 0, "WitnetRequestBoard: not yet solved");
```

contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -111,11 +111,13 @@ contract WitnetRequestBoard
         returns (bytes memory _bytecode)
     {
         WitnetTypes.DataRequest storage _dr = __dataRequest(id);
-        _bytecode = WitnetRequest(_dr.addr).bytecode();
-        require(
-            WitnetTypes.computeDataRequestCodehash(_bytecode) == _dr.codehash,
-            "WitnetRequestBoard: manipulated request"
-        );
+        if (_dr.addr != address(0)) {
```
tmpolaczyk, line 114: Let's add a comment explaining why this is needed:

The bridge node uses the `readDataRequest` function to check if the data request has been deleted. In that case, `readDataRequest` should return an empty array. This `if (_dr.addr != address(0))` is used to check the case where the data request has been deleted, because dr.addr is set to 0 in that case.

contracts/UsingWitnet.sol
```diff
@@ -3,26 +3,23 @@
...
-import "./Witnet.sol";
-import "./WitnetRequestBoardInterface.sol";
-
+import "./libs/Witnet.sol";
+import "./interfaces/WitnetRequestBoardInterface.sol";
 
 /**
  * @title The UsingWitnet contract
  * @notice Contract writers can inherit this contract in order to create Witnet data requests.
  */
```
lrubiorod, line 12:
```suggestion
 **/
```
There are several places where we use only one * and other we use **

mariocao (commented):
Some points to be discussed by the whole dev team, mostly regarding naming conventions:

- internal functions with starting with 2 underscores?
- arguments starting with 1 underscore?
- internal immutable variables with 2 underscores?
- interface names starting with `I`?
- abstract contracts names with prefix or suffix? (e.g. `...Base`)
- folders structure in general (e.g. data, impls, libs, utils,...)?
- struct names starting with `S`?
- code documentation guidelines (e.g. use `notice` and/or `dev`, use `\\\` vs. `/**/`)?

[gd]:
- renaming proposals:
  - `estimateGasCost(uint256)` to `estimateMinimalReward(uint256)`.
  - `requestsCount()` to either `getLastRequestId()` or `getNextRequestId()`.
  - `postDataRequest(..)` to `postRequest(..)`.
  - `upgradeDataRequest(uint256)` to `upgradeRequest(uint256)`.
  - `readDr` to `readRequest`.
  - `readDataRequest` to `readRequestBytecode`.
  - `readGasPrice` to `readRequestGasPrice`.
  - `readDrTxHash` to either `readResultTxHash` or `readWitnetTallyTxHash`.
  - eventually add `readWitnetTxEpochs(uint256)` returns commit, reveal and tally txs epoch numbers.
  - `github.com/witnet/witnet-ethereum-bridge` to `github.com/witnet/witnet-evm-oracle`.
- retyping proposals:
  - `WitnetData.Request.codehash` and `WitnetData.Request.txhash` as `bytes32`, instead of `uint256`.
  - `postRequest(address)` to `postRequest(WitnetRequest)`.
  - `SWitnetBoardDataRecord.result` as `WitnetData.Result`, instead of `bytes`.

contracts/UsingWitnet.sol
```diff
@@ -32,54 +29,61 @@ abstract contract UsingWitnet {
...
-  * @dev Call to `upgrade_dr` function in the WitnetRequestBoard contract.
-  * @param _id The unique identifier of a request that has been previously sent to the WitnetRequestBoard.
-  */
-  function witnetUpgradeRequest(uint256 _id) internal {
-    wrb.upgradeDataRequest{value: msg.value}(_id);
+  /**
+   * @notice Retrieves result of previously posted DR, and removes it from storage.
+   * @param _id The unique identifier of a previously posted data request.
+   * @return The result of the DR
+  **/
```
mariocao, line 48:
```suggestion
  */
```

contracts/WitnetProxy.sol
```diff
@@ -0,0 +1,97 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.6;
+
+/* solhint-disable avoid-low-level-calls */
+/* solhint-disable no-complex-fallback */
```
mariocao, line 6: Lint ignore rules should be set to the most restrictive level

contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -0,0 +1,319 @@
...
+            isUpgradable() && (                
+                _owner == address(0) ||
+                _owner == from
+            )
+        );
+    }
+
+    /// @dev Retrieves named version of current implementation.
+    function version() virtual public pure override returns (string memory) {
+        return "WitnetRequestBoard-Centralized-v0.3.1.3";
```
lrubiorod, line 79: I dont know if it is possible, but can it be read from a specific field?. My point is that if we update the contract but we miss update this function, it could provide wrong information

guidiaz: Yep, I agree.

Perhaps a better way to go, is to let the operator (aka: deployer) tag the new implementation that pretends to deploy and upgrade from the proxy. Moreover, as our migrations scripts run on javascript, they could even fetch the version tag from the `package.json`, for instance. 

A couple of things to have in mind, though: 

- Strings cannot (yet) be declared in Solidity as immutable fields, so it's necessary to change declaration of this method to:
  `function version() public pure override returns (bytes32)`. 
- Actual version tag should be provided on construction.

contracts/UsingWitnet.sol
```diff
@@ -32,54 +29,61 @@ abstract contract UsingWitnet {
...
-  * @param _id The unique identifier of a request that has been previously sent to the WitnetRequestBoard.
-  * @return A boolean telling if the request has been already resolved or not.
-  */
+  /**
+   * @notice Check if a request has been resolved by Witnet.
+   * @dev Contracts depending on Witnet should not start their main business logic (e.g. receiving value from third.
+   * parties) before this method returns `true`.
+   * @param _id The unique identifier of a request that has been previously sent to the WitnetRequestBoard.
+   * @return A boolean telling if the request has been already resolved or not.
+  **/
```
tmpolaczyk, line 38: What's the result of `witnetCheckRequestResolved` if the request has been resolved and then destroyed?

guidiaz: `false`.

Considering that only the `UsingWitnet` that posted the Witnet data request can actually destroy it, the `witnetCheckRequestResolved` should be interpreted as "check the given request hast already been solved by Witnet, but not yet destroyed by its requestor".

tmpolaczyk: I would add a comment saying that after destroying the result `witnetCheckRequestResolved` will return false.

contracts/WitnetProxy.sol
```diff
@@ -0,0 +1,102 @@
...
+    if (address(delegate) != address(0)) {
+      // New delegate address must differ from current one:
+      require(_newDelegate != address(delegate), "WitnetProxy: nothing to upgrade");
+
+      // Assert whether current implementation is intrinsically upgradable:
+      try delegate.isUpgradable() returns (bool _isUpgradable) {
+        require(_isUpgradable, "WitnetProxy: not upgradable");
+      } catch {
+        revert("WitnetProxy: unable to check upgradability");
+      }
```
tmpolaczyk, line 69: I think this check is redundant because `isUpgradableFrom` will fail if the contract is not upgradable, right?

```suggestion
```

contracts/UsingWitnet.sol
```diff
@@ -3,26 +3,23 @@
...
+  WitnetRequestBoardInterface internal immutable witnet;
 
  /**
   * @notice Include an address to specify the WitnetRequestBoard.
   * @param _wrb WitnetRequestBoard address.
-  */
+  **/
   constructor(address _wrb) {
-    wrb = WitnetRequestBoardInterface(_wrb);
+    require(_wrb != address(0), "UsingWitnet: null contract");
```
clbartoli, line 21: "WitnetRequestBoard: null contract"? and add one test for this require?

contracts/data/WitnetBoardDataACLs.sol
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @title Witnet Access Control Lists storage layout, for centralized request boards.
+ * @author Witnet Foundation
+ */
+abstract contract WitnetBoardDataACLs is WitnetBoardData {  
+  struct SWitnetBoardACLs {
+    mapping (address => bool) isReporter_;
+  }
+
+  constructor() {
```
clbartoli, line 16: maybe we could include in the constructor inputs the list of addresses that can be reporters, at the moment only the sender will be a reporter.  If there is going to be only one reporter with the onlyowner modifier should be enough

contracts/data/WitnetBoardDataACLs.sol
```diff
@@ -0,0 +1,32 @@
...
+    __acls().isReporter_[msg.sender] = true;
+  }
+
+  modifier onlyReporters {
+    require(__acls().isReporter_[msg.sender], "WitnetBoardDataACLs: unauthorized reporter");
+    _;
+  }  
+
+  function __acls() internal pure returns (SWitnetBoardACLs storage _struct) {
```
clbartoli, line 25: could you please add comments to this functions so that are easier to understand?

contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -0,0 +1,325 @@
...
+
+import "../WitnetBoard.sol";
+import "../../data/WitnetBoardDataACLs.sol";
+import "../../interfaces/WitnetRequestBoardInterface.sol";
+import "../../utils/Destructible.sol";
+import "../../utils/Upgradable.sol";
+
+/**
+ * @title Witnet Requests Board mocked
+ * @notice Contract to bridge requests to Witnet for testing purposes.
```
clbartoli, line 13:
```suggestion
 * @notice Contract to bridge requests to Witnet.
```

contracts/WitnetProxy.sol
guidiaz: It's intended to distinguish the case of a current implementation not being intrinsically upgradable (not even by the "owner"), from the case where a current implementation is potentially upgradable but the `msg.sender` is not authorized for performing the upgrade. In the first case a "not upgradable" revert message is produced, while "not authorized" in the second case.

contracts/UsingWitnet.sol
guidiaz: Great idea. I will add a test case within the `test/using_witnet.js` testing suite.

contracts/impls/centralized/WitnetRequestBoard.sol
```diff
@@ -0,0 +1,325 @@
...
+    /// @dev Retrieves the whole DR post record from the WRB.
+    /// @param id The unique identifier of a previously posted data request.
+    /// @return The DR record. Fails if DR current bytecode differs from the one it had when posted.
+    function readDr(uint256 id)
+        external view
+        virtual
+        wasPosted(id)
+        returns (WitnetData.Request memory)
+    {
+        return __checkDr(id);
```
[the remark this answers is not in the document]
guidiaz: Internal/private function calls are translated into simple JUMP evm-opcodes, so no extra gas is actually incurred.

contracts/data/WitnetBoardDataACLs.sol
guidiaz: The owner/operator can call `setReporters(address[])` at anytime. Besides, a list of reportes can be specified whenever a delegated WitnetRequestBoard instance is initialized (e.g. when upgrading and older implementation).

contracts/WitnetProxy.sol
```diff
@@ -0,0 +1,102 @@
...
+  /// @dev Constructor with no params as to ease eventual support of Singleton pattern (i.e. ERC-2470)
+  constructor () {}
+
+  /// @dev WitnetProxies will never accept direct transfer of ETHs.
+  receive() external payable {
+    revert("WitnetProxy: no ETH accepted");
+  }
+
+  /// @dev Payable fallback accepts delegating calls to payable functions.  
+  fallback() external payable { /* solhint-disable no-complex-fallback */
```
tmpolaczyk, line 26: When possible, use `solhint-disable-next-line` instead, I believe `solhint-disable` disables the lints for the rest of the current file:

```suggestion
  // solhint-disable-next-line no-complex-fallback
  fallback() external payable {
```

contracts/WitnetProxy.sol
```diff
@@ -0,0 +1,102 @@
...
+    address _delegate = address(delegate);
+
+    assembly { /* solhint-disable avoid-low-level-calls */
```
tmpolaczyk, line 29:
```suggestion
    assembly {
```

This lint is not needed here. Removing it results in some warnings on other lines, those can be disabled using `solhint-disable-next-line`

guidiaz:
Supersed by https://github.com/witnet/witnet-ethereum-bridge/pull/176

PR #181, opened 2021-08-18
Chore: Upgrade Solidity contracts to adopt WF code-style and naming conventions

### LINKED ISSUE #178: General upgrade of Solidity bridge contracts to adopt Witnet Foundation own code-style conventions

guidiaz:
Closes #178.


contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,497 @@
...
+
+
+    // ================================================================================================================
+    // --- Full implementation of 'IWitnetRequestBoardAdminACLs' -------------------------------------------------------------
+
+    /// Tells whether given address is included in the active reporters control list.
+    /// @param _reporter The address to be checked.
+    function isReporter(address _reporter) public view override returns (bool) {
+        return _acls().isReporter_[_reporter];
+    }
```
aesedepece, line 110: This whole reporters feature is over-engineered and deviates from the requirements, as we have never stated a need for more than one reporter. But hey, it works, so I have no problem with it for now.

aesedepece (changes requested):

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,497 @@
...
+
+    /// Requests the execution of the given Witnet Radon script in expectation that it will be relayed and solved by the Witnet DON.
+    /// A reward amount is escrowed by the Witnet Request Board that will be transfered to the reporter who relays back the Witnet-provided 
+    /// result to this request.
+    /// @dev Fails if:
+    /// @dev - provided reward is too low.
+    /// @dev - provided script is zero address.
+    /// @dev - provided script bytecode is empty.
+    /// @return _queryId The unique identifier of the posted request.
+    function postRequest(IWitnetRadon _script)
```
aesedepece, line 226: As mentioned before, it is untrue that this is a Radon script and we should avoid calling it like that. Let's agree on a better terminology for once, because this is getting more an more difficult hahaha

README.md
```diff
@@ -72,31 +72,31 @@ The `WitnetRequestBoard` implements the following functionality:
...
+- **`readRequestBytecode(uint256 _id)`**:
   - _Description_:
     - Retrieves the RADON script bytecode of a previously posted Witnet data request.
   - _Inputs_:
     - `_id`: the unique identifier of a previously posted Witnet data request.
   - _Returns_:
     - The Witnet RADON script bytecode, as `bytes`.
 
-- **`readResult(uint256 _id)`**:
+- **`readResponseWitnetResult(uint256 _id)`**:
```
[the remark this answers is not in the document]
guidiaz: There already exists a `readResponse` that returns the whole `WitnetData.Response` struct. I agree though with keeping `readResult` instead of the verbose `readResponseResult`.

contracts/UsingWitnet.sol
```diff
@@ -3,71 +3,75 @@
...
-  /// @notice Retrieves result of previously posted DR, and removes it from storage.
-  /// @param _id The unique identifier of a previously posted data request.
-  /// @return The result of the DR
-  function witnetDestroyResult(uint256 _id) internal returns (WitnetData.Result memory) {
-    return Witnet.resultFromCborBytes(_WRB.destroyResult(_id));
-  }
+    /// @notice Retrieves result of a previously posted request, and removes the whole query from the WRB's storage.
+    /// @param _id The unique identifier of a previously posted request.
+    /// @return The Witnet-provided result to the request.
+    function _witnetDestroyResult(uint256 _id) internal returns (Witnet.Result memory) {
```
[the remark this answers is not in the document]
guidiaz: Sounds good. Now it's a convenient moment to unify all naming conventions that we wish :-)

README.md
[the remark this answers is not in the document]
aesedepece: Ok,`readResponseResult` be it! This is opaque to `UsingWitnet` anyway.

contracts/impls/WitnetRadonBase.sol
```diff
@@ -0,0 +1,17 @@
...
+abstract contract WitnetRadonBase is IWitnetRadon {
+    using Witnet for bytes;
+    bytes public override bytecode;
+
+    /// Applies Witnet-compatible hash function over the `bytecode()` in order to 
+    /// uniquely identify every possible well-formed Radon script.
+    function codehash() external view override returns (bytes32) {
+      return bytecode.computeCodehash();
+    }
+}
```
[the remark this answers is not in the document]
guidiaz: I see ... however, something like `IWitnetDataRequestOutput` could also be misleading for developers/new comers, as it is mainly used as an "input" when posting a new "request" ... not having a better choice, what about something like **`IWitnetRequest`** ?

truffle-config.js
```diff
@@ -6,6 +6,7 @@ const realm = process.env.WITNET_EVM_REALM ? process.env.WITNET_EVM_REALM.toLowe
 
 module.exports = {
   build_directory: `./build/${realm}/`,
+  contractrs_directory: "./contracts/",
```
tmpolaczyk, line 9: I guess this is a typo?

```suggestion
  contracts_directory: "./contracts/",
```

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,497 @@
...
+        setReporters(abi.decode(_initData, (address[])));
+    }
+
+    /// Tells whether provided address could eventually upgrade the contract.
+    function isUpgradableFrom(address _from) external view override returns (bool) {
+        address _owner = _state().owner;
+        return (
+            // false if the WRB is intrinsically not upgradable
+            isUpgradable() && (                
+                _owner == address(0) ||
```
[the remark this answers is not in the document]
guidiaz: You are right. Indeed:

- `_owner == address(0)` was introduced as a way to return `true` in case the actual WRB instance (delegated by a proxy) had not yet been initialized, and because at that time there was no way to transfer owernship.

- As a `WitnetProxy` will never call this method before a new implementation is initialized, and as ownership could now be eventually transferred, indeed this condition MUST be removed.

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
guidiaz: Ok, sorry, it had a short and obvious implementation so I couldn't help it.

Future implementation, or this one, could eventually deprecate these kind of admin methods with no back-compatibility issues with user contracts, as these admin interfaces are in fact not inherited by the abstract `WitnetRequestBoard`.

truffle-config.js
guidiaz: Absolutely

contracts/impls/WitnetRadonBase.sol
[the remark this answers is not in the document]
guidiaz: So `IWitnetRequest` and `WitnetRequestBase` will be.

README.md
```diff
@@ -33,70 +34,77 @@ This table contains all the `WinetProxy` instances that act as entry-points for
...
+ The `IWitnetRequest` interface defines two methods:
 
 - **`bytecode()`**:
   - _Description_:
-    - Returns Witnet RADON script as a CBOR-encoded `bytes`. 
+    - Returns Witnet Data Request as a CBOR-encoded `bytes`. 
+
+- **`codehash()`**:
+  - _Description_:
+    - Returns Witnet Data Request as a CBOR-encoded `bytes`.
```
tmpolaczyk, line 51: This should say "Returns SHA256 hash of Witnet Data Request as a CBOR-encoded `bytes`"

contracts/libs/Witnet.sol
```diff
@@ -3,314 +3,370 @@
...
-  /// @param _result An instance of WitnetData.Result.
-  /// @return The `uint64[]` decoded from the WitnetData.Result.
-  function asUint64Array(WitnetData.Result memory _result) external pure returns(uint64[] memory) {
-    require(_result.success, "Tried to read `uint64[]` value from errored WitnetData.Result");
-    return _result.value.decodeUint64Array();
-  }
+    /// Data kept in EVM-storage containing Witnet-provided response metadata and result.
+    struct Response {
+        address reporter;       // Address from which the result was reported.
+        bytes32 proof;          // Witnet-provided validation proof of the reported result.
```
clbartoli, line 42: is this a proof or is it the transaction hash of the request? With proof is the PoI that comes to my mind

contracts/libs/Witnet.sol
```diff
@@ -3,314 +3,370 @@
...
-    if (_stageIndex == 0) {
-      return "retrieval";
-    } else if (_stageIndex == 1) {
-      return "aggregation";
-    } else if (_stageIndex == 2) {
-      return "tally";
-    } else {
-      return "unknown";
+    /// Data struct containing actual result data provided by the Witnet DON.
+    struct Result {
```
clbartoli, line 48: I'm not sure if it is necessary to change it, but it looks a bit confusing since only some lines above the result of the response is defined

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,499 @@
...
+            uint256 _queryId,
+            bytes32 _proof,
+            bytes calldata _result
+        )
+        external
+        override
+        onlyReporters
+        inStatus(_queryId, Witnet.QueryStatus.Posted)
+    {
+        _reportResult(_queryId, block.number, _proof, _result);
```
clbartoli, line 165: Since the epoch is needed for the data feed (but not the block number), we could set the epoch to 0 in this case, so that the data feeds knows that when is 0, but has proof != 0, has to look for the epoch using the timestamp of the transaction.  Otherwise it should detect somehow that the epoch provided is not a real epoch but a block number

contracts/libs/Witnet.sol
guidiaz: In general terms, it's meant to be a `bytes32` that enables the application to verify result correctness (and finality) of provided result within the context of the Witnet blockchain. In practice, in the centralized implementation: the hash used as "drTxHash" before.

contracts/libs/Witnet.sol
guidiaz: Indeed, `Response.result` is declared as `bytes` and `Result` contains the struct returned after passing these `bytes` to `WitnetParserLib.resultFromCborBytes()`. The reason for not declaring `Response.result` as `Witnet.Result` is because the WRB is/was not intended to perform any kind of parsing or decodification. 

I will add a more explicit comment to `struct Result`: "Data struct produced by the 'WitnetParserLib' when parsing a Witnet-provided raw result, encoded as bytes array."

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
guidiaz: The problem with that approach is that the "data feed [contract]" would have to guess the "timestamp" from the transaction that ultimately called to either `readResult` or `deleteQuery`, not from the one that called `reportResult`, which in all cases would be closer to the real and factual epoch in which the Data Request was actually solved by Witnet. 

`epoch` refers not to a timestamp, but to a block "height" in all cases.

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,501 @@
...
+        Payable,
+        IWitnetRequestBoardAdmin,
+        IWitnetRequestBoardAdminACLs,        
+        WitnetBoardDataACLs,
+        WitnetRequestBoardUpgradableBase        
+{
+    using Witnet for bytes;
+    using Witnet for Witnet.Result;
+
+    uint256 internal constant _ESTIMATED_REPORT_RESULT_GAS = 102496;
```
tmpolaczyk, line 28: According to the README, the gas cost is higher now:

```
|  Contract                            ·  Method          ·  Min        ·  Max        ·  Avg        ·  # calls     ·  usd (avg)  │
|  WitnetRequestBoardTrustableDefault  ·  reportResult    ·     118352  ·     120501  ·     118948  ·          18  ·          -  │
```

The `102496` constant comes from this commit: d29fd4284728dbdd6fa55ec074bc619f68e3ae55 and it is the "max", so the new value should be `120501`.

contracts/impls/trustable/WitnetRequestBoardTrustableOMGX.sol
```diff
@@ -0,0 +1,118 @@
...
+/// @author The Witnet Foundation
+contract WitnetRequestBoardTrustableOMGX
+    is
+        Payable,
+        WitnetRequestBoardTrustableBase
+{
+    uint256 internal lastBalance;
+    uint256 internal immutable _OVM_GAS_PRICE;
+
+    modifier ovmPayable {
```
clbartoli, line 27: Since its only used once, maybe the modifier is no needed. Anyhow, could you add some comment?

contracts/impls/trustable/WitnetRequestBoardTrustableDefault.sol
```diff
@@ -0,0 +1,72 @@
...
+    function destroy() external override onlyOwner {
+        selfdestruct(payable(msg.sender));
+    }
+
+
+    // ================================================================================================================
+    // --- Overrides 'Payable' ----------------------------------------------------------------------------------------
+
+    /// Gets balance of given address.
+    function balanceOf(address _from)
```
clbartoli, line 37: must this function be in this contract? I've seen it's only use in the WRBOMGX and maybe it can be defined there

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
guidiaz: Right. The slightly higher cost is mainly derived from the fact that `timestamp` and `epoch` are now stored for each newly reported result. Will then update the `_ESTIMATE_REPORT_RESULT_GAS` to `120501`. Should we not scare gas consumption, the `_reportResult` can be changed as to avoid writing these values to storage.

contracts/impls/trustable/WitnetRequestBoardTrustableOMGX.sol
guidiaz: In fact, the `upgradeQuery` should alse be overriden by imposing call to this modifier, so it's actually used twice in the same contract. Besides, it could be reused by an eventual new OMGX implementation inheriting from this one.

guidiaz: Comment added :-)

contracts/libs/Witnet.sol
guidiaz: Then again, if couple `WitnetParserLib` to the WRB, and make it expose the whole `IWitnetRequestParser` functionality (public methods of WitnetParserLib), we could totally decouple `UsingWitnet` (ergo PriceFeed contracts) from the `WitnetParserLib`, making developers life a little bit easier, and only one entry-point address to know about for each supported network.
Should the IWitnetRequestParser implementation be upgraded or patched, the upgrade would be done through out the Proxy, without forcing developers to deploy new instances of their contracts because of this.

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,694 @@
...
+        external view
+        override
+        returns (Witnet.QueryStatus)
+    {
+        return _getQueryStatus(_queryId);
+
+    }
+
+    /// Retrieves the whole Request record posted to the Witnet Request Board.
+    /// @dev Fails if the `_queryId` is not valid or, if it has been destroyed,
```
clbartoli, line 330:
```suggestion
    /// @dev Fails if the `_queryId` is not valid or, if it has been deleted,
```

clbartoli: this should be changed in every comment

.solhintignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 contracts/UsingWitnet.sol
+contracts/patterns/Payable.sol
 test/TestBuffer.sol
 test/TestActiveBridgeSet.sol
```
tmpolaczyk, line 5: Any reason to have these 4 lines here? I tried removing them and the only warning is about the `WitnetRequestSolved` modifier in `UsingWitnet.sol`, which could be renamed to `witnetRequestSolved`.

aesedepece:
How come `WitnetParserLib` not inherit from `IWitnetRequestParser`?

contracts/libs/Witnet.sol
```diff
@@ -3,314 +3,371 @@
...
-  /// @return The `uint64` decoded from the WitnetData.Result.
-  function asUint64(WitnetData.Result memory _result) external pure returns(uint64) {
-    require(_result.success, "Tried to read `uint64` value from errored WitnetData.Result");
-    return _result.value.decodeUint64();
-  }
+    /// Data kept in EVM-storage for every Request posted to the Witnet Request Board.
+    struct Request {
+        IWitnetRequest addr;    // The contract containing the Data Request which execution has been requested.
+        address requestor;      // Address from which the request was posted.
+        bytes32 codehash;       // Codehash of the Data Request which execution has been requested.
```
aesedepece, line 34: For simplicity:
```suggestion
        bytes32 hash;       // Hash of the Data Request whose execution has been requested.
```

On `whose` vs. `which`: https://thewritepractice.com/whose-vs-which-inanimate-objects/

contracts/libs/Witnet.sol
```diff
@@ -3,314 +3,371 @@
...
-  /// @param _result An instance of WitnetData.Result.
-  /// @return The `uint64` decoded from the WitnetData.Result.
-  function asUint64(WitnetData.Result memory _result) external pure returns(uint64) {
-    require(_result.success, "Tried to read `uint64` value from errored WitnetData.Result");
-    return _result.value.decodeUint64();
-  }
+    /// Data kept in EVM-storage for every Request posted to the Witnet Request Board.
+    struct Request {
+        IWitnetRequest addr;    // The contract containing the Data Request which execution has been requested.
+        address requestor;      // Address from which the request was posted.
```
aesedepece, line 33: Far more common naming for the person who requests:
```suggestion
        address requester;      // Address from which the request was posted.
```

contracts/libs/Witnet.sol
```diff
@@ -3,314 +3,371 @@
...
+        bytes32 codehash;       // Codehash of the Data Request which execution has been requested.
+        uint256 gasprice;       // Minimum gas price the DR resolver should pay on the solving tx.
+        uint256 reward;         // escrow reward to by paid to the DR resolver.
```
aesedepece, line 36: So as to match the "reporter" language down below:
```suggestion
        uint256 reward;         // Escrowed reward to by paid to the reporter of the Data Request result.
```

contracts/libs/Witnet.sol
```diff
@@ -3,314 +3,371 @@
...
-  /// @param _result An instance of WitnetData.Result.
-  /// @return The `uint64[]` decoded from the WitnetData.Result.
-  function asUint64Array(WitnetData.Result memory _result) external pure returns(uint64[] memory) {
-    require(_result.success, "Tried to read `uint64[]` value from errored WitnetData.Result");
-    return _result.value.decodeUint64Array();
-  }
+    /// Data kept in EVM-storage containing Witnet-provided response metadata and result.
+    struct Response {
+        address reporter;       // Address from which the result was reported.
+        uint256 timestamp;      // EVM-provided timestamp in which the result was reported.
```
aesedepece, line 42:
```suggestion
        uint256 timestamp;      // EVM-provided timestamp at which the result was reported. 
```

Because timestamps are specific times, much equivalent to stating hours and minutes: https://www.eurocentres.com/blog/talking-about-time-using-at-in-and-on

contracts/libs/Witnet.sol
aesedepece: The thing is that when the time comes to use an actual merkle proof here, it won't fit into `bytes32`, so it's a bit strange that we are thinking long-term but only half-way.

I can't imagine either any version of the request board that uses this field for anything else that fits into `bytes32`. I believe we can assume for certain that any future version of the request board that has a different trust model will necessarily require a new version of the `Response` structure that is laid out differently. If something, I can imagine that we may add a new field containing an actual proof, and the old field would be left empty or would be repurposed, so having this named `proof` may turn out to be confusing.

At this point, I think we are totally free to be explicit about what this represents. Otherwise we are just "hiding" the true nature of this field, right?

aesedepece (changes requested):

contracts/libs/Witnet.sol
tmpolaczyk: "to be paid"

README.md
```diff
@@ -262,29 +262,29 @@ Please, have a look at the [`witnet/truffle-box`](https://github.com/witnet/truf
...
 ·······································|·················|·············|·············|·············|··············|··············
-|  WitnetRequestBoardTrustableDefault  ·  postRequest    ·     144525  ·     196359  ·     164155  ·          33  ·          -  │
+|  WitnetRequestBoardTrustableDefault  ·  postRequest    ·     144650  ·     196484  ·     164280  ·          33  ·          -  │
 ·······································|·················|·············|·············|·············|··············|··············
-|  WitnetRequestBoardTrustableDefault  ·  reportResult   ·     119164  ·     121313  ·     119760  ·          18  ·          -  │
+|  WitnetRequestBoardTrustableDefault  ·  reportResult   ·     119210  ·     121359  ·     119806  ·          18  ·          -  │
 ·······································|·················|·············|·············|·············|··············|··············
-|  WitnetRequestBoardTrustableDefault  ·  upgradeReward  ·      31319  ·      36462  ·      34748  ·           6  ·          -  │
+|  WitnetRequestBoardTrustableDefault  ·  upgradeReward  ·      31341  ·      36484  ·      34770  ·           6  ·          -  │
```
aesedepece, line 277: I'm kind of missing `deleteQuery` benchmarked here. Do you think it makes sense to add it?

README.md
```diff
@@ -262,29 +262,29 @@ Please, have a look at the [`witnet/truffle-box`](https://github.com/witnet/truf
...
 |  WitnetDecoderLib                                      ·          -  ·          -  ·    1930600  ·      28.7 %  ·          -  │
 ·························································|·············|·············|·············|··············|··············
-|  WitnetParserLib                                       ·          -  ·          -  ·    2570010  ·      38.3 %  ·          -  │
+|  WitnetParserLib                                       ·          -  ·          -  ·    2594996  ·      38.6 %  ·          -  │
 ·························································|·············|·············|·············|··············|··············
 |  WitnetProxy                                           ·          -  ·          -  ·     587730  ·       8.7 %  ·          -  │
 ·························································|·············|·············|·············|··············|··············
-|  WitnetRequestBoardTrustableDefault                    ·          -  ·          -  ·    2487604  ·        37 %  ·          -  │
+|  WitnetRequestBoardTrustableDefault                    ·          -  ·          -  ·    3699603  ·      55.1 %  ·          -  │
```
aesedepece, line 287: How did this become +49% more expensive? :thinking:

aesedepece:
> > How come `WitnetParserLib` not inherit from `IWitnetRequestParser`?
> 
> At the time being, Solidity libraries cannot inherit, nor be inherited

Hahaha ok :+1:

contracts/data/WitnetBoardData.sol
```diff
@@ -2,73 +2,111 @@
...
-    internal view
-    returns (WitnetData.Query storage)
-  {
-    return _state().requests[_requestId].query;
-  }
-  
-  bytes32 internal constant _WITNET_BOARD_DATA_SLOTHASH =
-    /* keccak256("io.witnet.board.data") */
-    0x641d5bbf2c42118a382e660df7903a98dce7b5bb834d3ba9beae1890b2a72054;
+    modifier notDestroyed(uint256 _queryId) {
```
clbartoli, line 34: I think it should be `notDeleted`, and it should be changed in WRBTrustableBase.sol aswell

contracts/data/WitnetBoardData.sol
```diff
@@ -2,73 +2,111 @@
...
+        uint256 numQueries;
+        mapping (uint => Witnet.Query) queries;
+    }
+
+    constructor() {
+        _state().owner = msg.sender;
+    }
+
+
+    modifier inStatus(uint256 _queryId, Witnet.QueryStatus _status) {
```
clbartoli, line 27: could you please add a small comment in the modifiers? :)

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,694 @@
...
+        }
+        emit ReportersUnset(_exReporters);
+    }
+
+
+    // ================================================================================================================
+    // --- Full implementation of 'IWitnetRequestBoardReporter' -------------------------------------------------------
+
+    /// Reports the Witnet-provided result to a previously posted request. 
+    /// @dev Will assume `block.number` as the epoch number for the provided result.
```
tmpolaczyk, line 148: Is this comment true? `epoch` is set to 0, and I don't see any mentions of `block.number`.

contracts/interfaces/IWitnetRequestBoardReporter.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.7.0 <0.9.0;
+
+/// @title The Witnet Request Board Reporter interface.
+/// @author The Witnet Foundation.
+interface IWitnetRequestBoardReporter {
+    /// Reports the Witnet-provided result to a previously posted request. 
+    /// @dev Will assume `block.number` as the epoch number for the provided result.
```
tmpolaczyk, line 9: Same comment, verify if it actually uses the `block.number`.

README.md
```diff
@@ -32,132 +34,201 @@ This table contains all the `WinetProxy` instances that act as entry-points for
...
-- **`destroyResult(uint256 _id)`**:
+- **`upgradeReward(uint256 _id)`**:
+  - _Description_: increments the reward of a Witnet data request by 
+  adding more value to it. The new data request reward will be increased by `msg.value`.
+  - _Inputs_:
+    - `_id`: the unique identifier of a previously posted Witnet data request.
+
+- **`deleteQuery(uint256 _id)`**:
   - _Description_:
     - Retrieves the Witnet-solved result (if already available) of a previously posted Witnet request, and removes it from the WRB storage.
```
clbartoli, line 99: This function removes the entire request, doesn't it? In this comment looks like only the result is deleted

contracts/libs/Witnet.sol
guidiaz: Certain as we are that an eventual merkle proof would not fit within `bytes32` field, let's better name it as what it is in practive: "witnetTxHash", or simply "txhash" ?

.solhintignore
guidiaz: Not any more, really. Most of the lines were due to previous previous. Payable can be also removed after last modification attending one of cbartoli's comments

contracts/libs/Witnet.sol
```diff
@@ -3,314 +3,371 @@
...
-  /// @notice Decode an error code from a WitnetData.Result as a member of `WitnetData.ErrorCodes`.
-  /// @param _result An instance of `WitnetData.Result`.
-  /// @return The `CBORValue.Error memory` decoded from the WitnetData.Result.
-  function asErrorCode(WitnetData.Result memory _result) external pure returns(WitnetData.ErrorCodes) {
-    uint64[] memory error = asRawError(_result);
-    if (error.length == 0) {
-      return WitnetData.ErrorCodes.Unknown;
+    /// @notice Witnet lambda function that computes the hash of a CBOR-encoded Data Request.
+    /// @param _bytecode CBOR-encoded RADON.
+    function computeCodehash(bytes memory _bytecode) internal pure returns (bytes32) {
```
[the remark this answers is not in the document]
guidiaz: perhaps, would it be better to be explict and name it as `sha256()` ?

contracts/libs/Witnet.sol
guidiaz: Thanks for the grammar tip :-) 👍

contracts/libs/Witnet.sol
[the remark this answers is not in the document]
guidiaz: `hash(bytes)` as `sha256(bytes)` would shadow a bultin symbol.

README.md
guidiaz: Will double-check, but should mainly be because of:

- new getter methods declared in `IWitnetRequestBoardView`
- new getter methods declared in `IWitnetRequestParser`
- new methods in `IWitnetRequestBoardAdminACLs`

contracts/data/WitnetBoardData.sol
guidiaz: Right you are :-)

contracts/data/WitnetBoardData.sol
guidiaz: Yeap, sorry again

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
guidiaz: Sorry, not anymore. @dev comment removed

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
guidiaz: Better still, instead of `epoch` let reporters directly specify the Witnet-provided timestamp, in which case `Response.epoch` would be no longer needed.

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,694 @@
...
+    function readResponse(uint256 _queryId)
+        external view
+        override
+        inStatus(_queryId, Witnet.QueryStatus.Reported)
+        returns (Witnet.Response memory _response)
+    {
+        return _getResponseData(_queryId);
+    }
+
+    /// Retrieves the hash of the Witnet transaction hash that actually solved the referred query.
```
aesedepece, line 405:
```suggestion
    /// Retrieves the hash of the Witnet transaction that actually solved the referred query.
```

contracts/interfaces/IWitnetRequestBoardRequestor.sol
```diff
@@ -0,0 +1,42 @@
...
+
+/// @title Witnet Requestor Interface
+/// @notice It defines how to interact with the Witnet Request Board in order to:
+///   - request the execution of Witnet Radon scripts (data request);
+///   - upgrade the resolution reward of any previously posted request, in case gas price raises in mainnet;
+///   - read the result of any previously posted request, eventually reported by the Witnet DON.
+///   - remove from storage all data related to past and solved data requests, and results.
+/// @author The Witnet Foundation.
+interface IWitnetRequestBoardRequestor {
+    /// Retrieves copy of all Witnet-provided data related to a previously posted request, removing the whole query from the WRB storage.
```
aesedepece, line 16:
```suggestion
    /// Retrieves a copy of all Witnet-provided data related to a previously posted request, removing the whole query from the WRB storage.
```

aesedepece (changes requested):

README.md
aesedepece: Cool. It's deployment cost anyway, so not a big concern. Thanks! :smiley:

package.json
```diff
@@ -12,22 +12,21 @@
...
-    "flatten:witnet": "npm run clean && npm run flatten:witnet:libs && npm run flatten:witnet:proxy && npm run flatten:witnet:board",
-    "flatten:witnet:libs": "npm run flatten contracts/libs/Witnet.sol && npm run flatten contracts/libs/CBOR.sol",
-    "flatten:witnet:proxy": "npm run flatten contracts/WitnetProxy.sol",
-    "flatten:witnet:board": "npm run flatten:witnet:board:trustable && npm run flatten:witnet:board:trustless",
-    "flatten:witnet:board:trustable": "npm run flatten contracts/impls/trustable/",
-    "flatten:witnet:board:trustless": "npm run flatten contracts/impls/trustless/",
+    "flatten:witnet": "npm run clean && npm run flatten:witnet:libs && npm run flatten:witnet:proxy && npm run flatten:witnet:boards",
+    "flatten:witnet:libs": "npm run flatten contracts/libs/WitnetDecoderLib.sol && npm run flatten contracts/libs/WitnetParserLib.sol",
+    "flatten:witnet:proxy": "npm run flatten contracts/impls/WitnetProxy.sol",
+    "flatten:witnet:boards": "npm run flatten:witnet:boards:trustable && npm run flatten:witnet:boards:trustless",
```
clbartoli, line 18: There is no trustless folder, so this command fails

package.json
```diff
@@ -12,22 +12,21 @@
...
+    "flatten:witnet:boards:trustable": "npm run flatten contracts/impls/trustable/",
+    "flatten:witnet:boards:trustless": "npm run flatten contracts/impls/trustless/",
```
clbartoli, line 20: same here, it should be deleted and added in the future if necessary

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
tmpolaczyk: Also, currently this constant is unique for all the different implementations. Maybe it should be defined somewhere else to allow supporting other EVM-compatible chains with different gas costs.

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -0,0 +1,694 @@
...
+import "../../interfaces/IWitnetRequestBoardAdminACLs.sol";
+import "../../libs/WitnetParserLib.sol";
+import "../../patterns/Payable.sol";
+
+/// @title Witnet Request Board "trustable" base implementation contract.
+/// @notice Contract to bridge requests to Witnet Decentralized Oracle Network.
+/// @dev This contract enables posting requests that Witnet bridges will insert into the Witnet network.
+/// The result of the requests will be posted back to this contract by the bridge nodes too.
+/// @author The Witnet Foundation
+abstract contract WitnetRequestBoardTrustableBase
```
clbartoli, line 18: I think this contract should inherit `WitnetRequestBoard.sol` as to make sure before deploying  it that  it contains all the functions needed, or am I missing something?

clbartoli: I mean before deploying WRBTDefault and WRBTBoba

package.json
guidiaz: Right, as empty folders are not commited to github

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
guidiaz: `WitnetRequestBoardTrustableBase` inherits from `WitnetRequestBoardUpgradableBase` from which WitnetRequestBoard is inherited. So in fact, `WitnetRequestBoardTrustableBase` actually inherits from `WitnetRequestBoard` :-)

aesedepece (approved):
carita feliz

clbartoli (approved):

PR #201, opened 2021-09-13
Feat: Infer realm from network name passed by args

guidiaz:
- Try to read `WITNET_EVM_REALM` env variable only if no `--network`  value is passed from command line. 
- Defaults to "default" if `WITNET_EVM_REALM` is not declared either. 
- Networks other than"development" or "test", will be looked upon the `./migrations/winet.settings.js` configuration file.


aesedepece:
Let's take this chance to get the final addresses and networks structures exported as JavaScript modules so that they can be imported into other repositories without having to do anything exotic.

guidiaz:
> Let's take this chance to get the final addresses and networks structures exported as JavaScript modules so that they can be imported into other repositories without having to do anything exotic.

Converted `migrations/witnet.addresses.json` into `migrations/witnet.addresses.js`.

scripts/utils.js
```diff
@@ -0,0 +1,35 @@
...
+  let network = "development"
+  const args = process.argv.join("=").split("=")
+  const networkIndex = args.indexOf("--network")
+  if (networkIndex >= 0) {
+    network = args[networkIndex + 1]
+    realm = (network || "default").split(".")[0]
+  }
+  if (realm === "ethereum") realm = "default"
+  return [realm, network]
+}
```
aesedepece, line 21: For more consistency, unified handling of defaults and reduced code repetition:

```suggestion
function getRealmNetworkFromArgs () {
  // Try to extract realm/network info from environment
  let network = process.env.WITNET_EVM_REALM
    ? process.env.WITNET_EVM_REALM.toLowerCase()
    : "default"
  // If a `--network` argument is provided, use that instead
  const args = process.argv.join("=").split("=")
  const networkIndex = args.indexOf("--network")
  if (networkIndex >= 0) {
    network = args[networkIndex + 1]
  }

  // Tell realm and network from one another
  return getRealmNetworkFromNetwork(network)
}
```

This is a common pattern that I really love: allowing the impure function (`getRealmNetworkFromArgs`) to only handle the side-effects and then delegate the functional business logic to a pure function (`getRealmNetworkFromNetwork`).

aesedepece (changes requested):

scripts/utils.js
guidiaz: Nice. Moreover, there was still some logic missing: 
- What if WITNET_EVM_REALM is set and no prefix in network string? Add WITNET_EVM_REALM value as prefix. 
- What if WITNET_EVM_REALM is set, but network string refers to a different realm? Catch and throw.

scripts/utils.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+  } else {
+    realm = envRealm || "default"
+    network = `${realm === "default" ? "ethereum" : realm}.${network}`
+  }
+  if (realm === "default") {
+    const subnetwork = network.split(".")[1]
+    if (subnetwork === "development" || subnetwork === "test") {
+      // Networks in realm "default", other than "developtment" or "test",
+      // must be prefixed with "ethereum."."
```
aesedepece, line 50: I'm unsure about the accuracy of this comment :thinking:

guidiaz: Changed to: `In "default" realm, networks "development" and "test" must be returned without a prefix.`

aesedepece (approved):

PR #209, opened 2021-11-05
Feat: Add reference implementation of a Witnet Randomness request contract

### LINKED ISSUE #208: Add Witnet RNG request contract
This contract should have the RNG bytecode but it should allow to configure the number of:
- witnessing nodes
- reward
- collateral

guidiaz:
Closes #208.


contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,234 @@
...
+    /// @dev Modifier: makes sure latest request was already solved.
+    modifier notPending {
+        require(isReady(), "WitnetRNG: pending randomize");
+        _;
+    }
+
+    /// Gets randomness generated upon latest request.
+    /// @dev Returns 0x0...0 if not yet solved, or 0xf...f if randomness could not get solved by Witnet for any 
+    /// @dev unexpected reason. Fails if `randomize()` was never called before.
+    function getRandomness()
```
tmpolaczyk, line 33: I think it would be a better API to have this function revert in case of error, because it is a public function so any user may call it and forget to check this special conditions.

guidiaz: The issue would then be how to enable clients in general, and the contract owner in particular, a simple way to detect the weird case in which the WitnetRequestRandomness resolves with an error (i.e. 0xf...f). Perhaps turning @dev comment as part of the @notice description would help unaware developers.

With respect letting 0x0 as returned value when no randomize() was ever called, seems fair enough to me.

contracts/patterns/Clonable.sol
```diff
@@ -0,0 +1,48 @@
...
+        public virtual
+        returns (Clonable _instance)
+    {
+        address _self = __self;
+        assembly {
+            let ptr := mload(0x40)
+            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
+            mstore(add(ptr, 0x14), shl(0x60, _self))
+            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
+            _instance := create(0, ptr, 0x37)
```
lrubiorod, line 24: Could you add a comment to this?

aesedepece:
This sort of differs quite much from what I had in mind for #208, but looks really interesting anyway.

In my mind, #208 was all about creating a contract that saves the user the need for using `rad2sol` if they are using only the RNG functionality. That is, providing a pre-built RNG data request that can be easily instantiated and posted through the usual `postDataRequest` method.

This however looks more like a factory for recurrently creating RNG requests, so it goes after a higher-level functionality that is more or less in the level of abstraction of the pre-built price feed contracts.

These two approaches are complementary, and I believe there's a case for both. But let's make sure that both exist!

Regarding the design of this specific solution, it looks quite well thought through, and I only have two major objections:

1. Regenerating the entire bytecode for a request every time that one of the `set*` functions is used seems really wasteful to me, specially under the assumption that for any given requests, the user may want to customize many of these fields. I'd rather go for a "builder" pattern in which the setters simply put the values into a structure, and only once we need the bytecode, we call a function that does the `abi.encodePacked` magic and puts the resulting bytes in the right place.

2. This has a very opinionated trust model, inasmuch it introduces `onlyOwner` and essentially allows the requester to deny execution of a data request at will. This is a non-trivial assumption that we have extensively run off of for a long time, so this needs to be discussed and assessed very carefully.

On a more fundamental point, I would like to know if you see this as being deployed and operated by the user (1 instance of this will be serving multiple RNG requests to their contracts over time), or whether you envision this as a sort of singleton like the request board or the price feeds, which serve multiple users at once.

aesedepece:
Ok, this slide answers my last question:

image

guidiaz:
> 1. Regenerating the entire bytecode for a request every time that one of the `set*` functions is used seems really wasteful to me, specially under the assumption that for any given requests, the user may want to customize many of these fields. I'd rather go for a "builder" pattern in which the setters simply put the values into a structure, and only once we need the bytecode, we call a function that does the `abi.encodePacked` magic and puts the resulting bytes in the right place.

Concatenating the whole bytecode every time `bytecode()` or `hash()` is called, would make state-modifying tx calling these methods more expensive in terms of used gas. For instance, `postRequets(IWitnetRequest)` on the WitnetRequestBoard implementation. Current design (in which one single SLOAD is required on either `bytecode()` and `hash()`, was preferrably followed on the assumption that `set*` methods are not expected to be so frequently called, they can only be called (and paid for) by the owner address, and they may eventually become uncallable (i.e. if ownership is renounced),

contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /// Requests the Witnet oracle to generate an EVM-agnostic and trustless source of randomness. 
+    /// Only one randomness request per block will be actually posted to the WRB. Should there 
+    /// already be a posted request within current block, all received funds shall be transfered
+    /// back to the tx sender.
+    function randomize()
+        external payable
+        virtual override
+        returns (uint256 _id)
```
tmpolaczyk, line 193: It says that this function returns an `_id`, but the documentation does not explain what is this `_id` and what to do with it.

I expected this to be a "randomness query id", such that the client will need this id to get their random number later. However that's not the case, the client must use a "uint256 block" in order to get the randomness. I assume that's the ethereum block number but it's also not clear by reading the documentation. Should that be the same block number as the one from the call to randomize? So I'm missing some documentation on how to use this contract.

Maybe it will all be explained in a high level tutorial somewhere else, but by reading the documentation of this file, if I were creating a solidity app I wouldn't know how to integrate this contract safely.

contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,338 @@
...
+        );
+    }
+
+    /// @dev Recursively searches for first block before the given one in which a Witnet Randomness request was casted.
+    /// @dev Returns 0 if none found.
+
+    function _searchPrevBlock(uint256 _target, uint256 _latest) internal view returns (uint256) {
+        return ((_target > _latest)
+            ? _latest
+            : _searchPrevBlock(_target, __randomize_[_latest].prevBlock)
```
tmpolaczyk, line 335: Let's hope that solidity optimizes away the recursion :crossed_fingers:

In some cases this kind of "search" functions can be optimized by providing a "hint" argument, and calculating that "hint" off-chain. This can save some gas in the worst case scenarios, but checking that additional argument also may cost some gas. And also the user contract must be modified to support it, which adds complexity, and I think this contract is already complex enough...

contracts/examples/WitnetRNG.sol
guidiaz: Indeed, returning the queryId as it doing makes no sense, as there are no methods in the interface accepting this queryId as a parameter. It could return block.number instead, but this is something the calling contract already knows so there's no actual need either to return this value. So yep, it seems better not to return any value at all from randomize().

contracts/examples/WitnetRNG.sol
guidiaz: The recursion call is within internal methods, so user contract would not need to deal with them at all. Concerning recursion optimization, proposed algorithm requires reading one word, and only one, per recursion step (what's to say ~200 gu per iteration).

guidiaz:
There are two major approaches to request randomness from the Witnet oracle:

**1.** Posting a **WitnetRequestRandomness** instance to the WitnetRequestBoard entrypoint:
  - **Pros**:
    - You, or your contract, will have full low-level control of the request update flow.
    - WitnetRandomnessRequest is an Ownable contract, letting owner address modify actual bytecode. Ownership can be renounced at any time, in which moment the bytecode would become totally immutable. 
    - While ownership is not renounced, you, or your contract, will be able to alter any witnessing parameter of the actual data request that will get posted to the Witnet oracle on the next update request.
  - **Cons**:
    - WitnetRandomnessRequest is an Ownable contract, letting owner address modify actual bytecode. Ownership can be renounced at any time, in which moment the bytecode would become totally immutable. 
    - Deploying a WitnetRandomnessRequest from scratch requires >2M gas units. You may overcome this handicap by "cloning" an existing WitnetRandomnessRequest (~200K gas units). Ownership of cloned WitnetRandomnessRequest will be granted to the cloner.
    - You, or your contract, will have to pay the Witnet fee everytime a new randomness request needs to be posted.
    - You, or your contract, will have to implement the logic to check whether the latest randomness request has either been already solved, or not, or handle extreme cases in which Witnet is not capable to resolve the request (for instance, if the witnessing parameters of the WitnetRandomnessRequest get modified before a previously posted request get resolved).
    - In you needed to request additional randomness before a previous one gets solved, you, or your contract, will have to implement the logic and data model to solve this functionality.
    
    - 2. High-level: using

guidiaz:
**WitnetRequestRandomness** example code snippets:
```
import "witnet-solidity-bridge/contracts/UsingWitnet.sol";
import "witnet-solidity-bridge/contracts/requests/WitnetRandomnessRequest.sol";

contract MyContract is UsingWitnet {

    WitnetRandomnessRequest witnetRandomnessRequest;
    uint internal __latestRandomnessRequestId;

    constructor(WitnetRequestBoard _wrb) UsingWitnet(_wrb) {
        witnetRandomnessRequest = new WitnetRandomnessRequest();
    }

    function pendingRandomness() public view returns (bool) {
        return !_witnetCheckResultAvailability(__latestRandomnessRequestId);
    }

    function randomize() external payable {
        if (!pendingRandomness()) {
            uint _usedFunds;
            (__latestRandomnessRequestId, _usedFunds) = _witnetPostRequest(witnetRandomnessRequest);
            payable(msg.sender).transfer(msg.value - _usedFunds);
        }
    }
    

}
```

guidiaz:
**WitnetRNG** example code snippets:

- One-phase randomness game:
  ```solidity
  ```
 
- Two-phase randomness game:
  ```solidity
  ```

- Three-phase randomness game:
  ```solidity
  ```

guidiaz:
- ***Two-phase randomness game** (Commit/Claim):*
  ```solidity
  import "witnet-solidity-bridge/contracts/interfaces/IWitnetRNG.sol"

  contract MyContract {

      uint256 public jackpot;
      uint32 public immutable oddity;
      IWitnetRNG public immutable witnetRNG;
      mapping (address => uint256) public balanceOf;

      struct Wager {
          address bidder;
          bool claimed;
      }
      mapping (uint256 => Wager) internal __wagers;

      constructor(uint32 _oddity) {
          assert(_oddity > 0);
          oddity = _oddity;
          witnetRNG = IWitnetRNG("<address of the WitnetRNG instance>");
      }

      receive() external payable {
          jackpot += msg.value;
      }

      function play() external payable {
          require(__wagers[block.number].bidder == address(0), "already played");
          __wagers[block.number].bidder = msg.sender;

          uint _fee = witnetRNG.estimateRandomizeFee();
          require(msg.value >= 10 * _fee, "not enough credit");

          uint _actualFee = witnetRNG.randomize();
          jackpot += (msg.value - _actualFee);
      }

      function winnerOf(uint256 _blockNumber) public view returns (address _winner) {
          address _bidder = __wagers[_blockNumber].bidder;
          if (_bidder != address(0) && **witnetRNG.isRandomized**(_winner)) {
              if (_isWinner(_blockNumber)) {
                  winner = _bidder;
              }
          }
      }

      function claim(uint _blockNumber) external {
          Wager storage __wager = __wagers[_blockNumber];
          if (
              __wager.bidder != address(0)
                  && __wager.claimed == false
                  && _isWinner(_blockNumber)
          ) {
              __wager.claimed = true;

              uint _prize = jackpot / 2;  // (ever-growing jackpot)
              balanceOf[__wager.bidder] += _prize;
              jackpot -= _prize;
          }
      }

      function withdraw() external {
          uint _totalPrize = balanceOf[msg.sender];
          if (_totalPrize > 0) {
              balanceOf[msg.sender] = 0;
              payable(msg.sender).transfer(_totalPrize);
          }
      }

      function _isWinner(uint _blockNumber) internal view returns (bool) {
          return witnetRNG.random(oddity, 0, _blockNumber) == 0;
      }
    
  }

guidiaz:
- ***Three-phase randomness game** (Commit/Reveal/Claim):*
  ```solidity
  
  ```

contracts/examples/WitnetRNG.sol
```diff
@@ -91,8 +91,13 @@ contract WitnetRNG
...
         Witnet.Result memory _witnetResult = _witnetReadResult(_queryId);
-        require(witnet.isOk(_witnetResult), "WitnetRNG: failed randomize");
-        return witnet.asBytes32(_witnetResult);
+        if (witnet.isOk(_witnetResult)) {
+            return witnet.asBytes32(_witnetResult);
+        } else {
+            uint256 _nextRandomizeBlock = __randomize_[_block].nextBlock;
+            require(_nextRandomizeBlock != 0, "WitnetRNG: faulty randomize");
+            return getRandomnessAfter(_nextRandomizeBlock);
+        }
```
tmpolaczyk, line 100: Interesting.

The documentation of `getRandomnessAfter` needs to be updated as well.

guidiaz: So right. Done

contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,352 @@
...
+            ) & uint256(2 ** _flagBits - 1);
+        return uint32((_number * _range) >> _flagBits);
+    }
+
+    /// Requests the Witnet oracle to generate an EVM-agnostic and trustless source of randomness. 
+    /// Only one randomness request per block will be actually posted to the WRB. Should there 
+    /// already be a posted request within current block, it will try to upgrade Witnet fee of current's 
+    /// block randomness request according to current gas price. In both cases, all unused funds shall 
+    /// be transfered back to the tx sender.
+    /// @return _usedFunds Amount of funds actually used from those provided by the tx sender.
```
aesedepece, line 199:
```suggestion
    /// Requests the Witnet oracle to generate an EVM-agnostic and trustless randomness source.
    /// Only one randomization request per block will be actually posted to the WRB. Should there
    /// already be a posted request within current block, it will try to upgrade Witnet fee of current's
    /// block randomization request according to current gas price. In both cases, all unused funds
    /// shall be transferred back to the transaction sender.
    /// @return _usedFunds Amount of funds actually used from those provided by the tx sender.
```

contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,352 @@
...
+        } else {
+            return upgradeRandomizeFee(block.number);
+        }
+    }
+
+    /// Increases Witnet fee related to a pending-to-be-solved randomness request, as much as it
+    /// may be required in proportion to how much bigger the current tx gas price is with respect the 
+    /// highest gas price that was paid in either previous fee upgrades, or when the given randomness 
+    /// request was posted. All unused funds shall be transferred back to the tx sender.
+    /// @return _usedFunds Amount of dunds actually used from those provided by the tx sender.
```
aesedepece, line 238:
```suggestion
    /// Increases Witnet fee related to a pending-to-be-solved randomization request, as much as it
    /// may be required in proportion to how bigger the current transaction gas price is than the
    /// highest gas price that was paid in either previous fee upgrades, or when the given randomization 
    /// request was posted. All unused funds shall be transferred back to the transaction sender.
    /// @return _usedFunds Amount of funds actually used from those provided by the transaction sender.
```

contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,352 @@
...
+        _v |= _v >> 4;
+        _v |= _v >> 8;
+        _v |= _v >> 16;
+        return _bitPosition[
+            uint32(_v * uint256(0x07c4acdd)) >> 27
+        ];
+    }
+
+    /// @dev Recursively searches for first block after the given one in which a Witnet Randomness request was casted.
+    /// @dev Returns 0 if none found.
```
aesedepece, line 335:
```suggestion
    /// @dev Recursively searches for the number of the first block after the given one in which a Witnet randomization request was posted.
    /// @dev Returns 0 if none found.
```

contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,352 @@
...
+    function _searchNextBlock(uint256 _target, uint256 _latest) internal view returns (uint256) {
+        return ((_target >= _latest) 
+            ? __randomize_[_latest].nextBlock
+            : _searchNextBlock(_target, __randomize_[_latest].prevBlock)
+        );
+    }
+
+    /// @dev Recursively searches for first block before the given one in which a Witnet Randomness request was casted.
+    /// @dev Returns 0 if none found.
+
```
aesedepece, line 345:
```suggestion
    /// @dev Recursively searches for the number of the first block before the given one in which a Witnet randomization request was posted.
    /// @dev Returns 0 if none found.
```

contracts/interfaces/IWitnetRNG.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.7.0 <0.9.0;
+
+import "../libs/Witnet.sol";
+
+/// @title The Witnet Random Number Generation provider interface.
+/// @author The Witnet Foundation.
+interface IWitnetRNG {
```
aesedepece, line 9: From a marketing perspective, we'd better forget about RNG and focus on _randomness_. This could be `IWitnetRandom` or `IWitnetRandomness`

aesedepece: For comments on interface docs, see comments on implementation.

test/TestWitnetDecoderLib.sol
```diff
@@ -121,6 +121,18 @@ contract TestWitnetDecoderLib {
...
+    // TODO
+  }
+
+  function testBytes32DecodeValueFrom32bytes() external {
+    // TODO
+  }
+
+  function testBytes32DecodeValueFrom33bytes() external {
+    // TODO
+  }
```
aesedepece, line 134: ???

contracts/requests/WitnetRequestMalleableBase.sol
```diff
@@ -0,0 +1,374 @@
...
+        _request().hash = _request().bytecode.hash();
+        emit WitnessingParamsChanged(
+            msg.sender,
+            _numWitnesses,
+            _minWitnessingConsensus,
+            _witnessingCollateral,
+            _witnessingReward,
+            _witnessingUnitaryFee
+        );
+    }
```
aesedepece, line 319: As I originally commented on this piece of code, it seems rather overkill to me to serialize the whole thing only to malleate a single field.

contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,352 @@
...
+    /// @param _wrb The WitnetRequestBoard immutable entrypoint address.
+    constructor(WitnetRequestBoard _wrb)
+        UsingWitnet(_wrb)
+    {
+        witnetRandomnessRequest = new WitnetRequestRandomness();
+        witnetRandomnessRequest.transferOwnership(msg.sender);
+    }
+
+    /// Returns amount of weis required to be paid as a fee when requesting randomness with a 
+    /// tx gas price as the one given.
```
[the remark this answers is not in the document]
aesedepece: All my comments on _randomness_ vs _randomization_ come from the fact that this contract is designed not to serve randomness directly but rather randomize / reseed itself, and then deterministically derive random stuff from that. My hunch is that _randomization_ better describes what's going on and avoids misconceptions on what the contract actually does.

aesedepece (changes requested):

contracts/examples/WitnetRNG.sol
```diff
@@ -0,0 +1,352 @@
...
+
+pragma solidity >=0.7.0 <0.9.0;
+pragma experimental ABIEncoderV2;
+
+import "../UsingWitnet.sol";
+import "../interfaces/IWitnetRNG.sol";
+import "../requests/WitnetRequestRandomness.sol";
+
+/// @title WitnetRNG: A trustless random number generator and registry, based on the Witnet oracle. 
+/// @author The Witnet Foundation.
```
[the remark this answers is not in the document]
guidiaz: a registry that keeps track of all randomness served in the past

contracts/examples/WitnetRNG.sol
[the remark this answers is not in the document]
guidiaz: i had a similar feeling while making progress on the development _ i fully agree on relying only on the "randomization" concept

test/TestWitnetDecoderLib.sol
guidiaz: My bad 🙈

contracts/requests/WitnetRequestMalleableBase.sol
guidiaz: `_malleateBytecode(..)` can be called from three public functions: `setWitnessingCollateral(uint64)`, `setWitnessingFees(uint64,uint64)` and `setWitnessingQuorum(uint8,uint8)`. The latter two actually modify two fields, not just one. 

I agree first thought would be keep track of offset positions in which each field is placed, but it's not that simple because length of "varint" values depend on the actual value to be encoded. Checking actual length before overwritting would not preserve the need of re-concatenating the whole bytecode.

test/TestWitnetDecoderLib.sol
aesedepece: This is my only objection!

guidiaz: Now fixed in https://github.com/witnet/witnet-solidity-bridge/pull/209/commits/b2c8881c371f855a0943a361f1dafeabe1253497.

aesedepece (approved):

PR #216, opened 2021-12-23
Feat: Optimize reportResult's required gas

guidiaz:


test/wrb.js
```diff
@@ -359,20 +359,6 @@ contract("WitnetRequestBoard", ([
...
-
-      // Should revert when reading the request since the bytecode has changed
-      await expectRevert(
-        this.WitnetRequestBoard.readRequestBytecode(queryId, {
-          from: other,
-          gasPrice: 1,
-        }),
-        "bytecode changed after posting"
-      )
-    })
```
aesedepece, line 375: Why is this being removed?

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -694,16 +694,18 @@ abstract contract WitnetRequestBoardTrustableBase
...
+        __response.timestamp = _timestamp;
+        __response.drTxHash = _drTxHash;
+        __response.reporter = msg.sender;
+        __response.cborBytes = _cborBytes;
 
-        _safeTransferTo(payable(msg.sender), _query.request.reward);
+        _safeTransferTo(payable(msg.sender), __query.request.reward);
         emit PostedResult(_queryId, msg.sender);
+
+        delete __query.request;
```
aesedepece, line 709: This is the core change of this PR. Am I understanding correctly that from now, in every single case, the `request` part of queries will be dropped as soon as we have a result reported?

This is interesting. I can't really think of an use case at this point for archiving the requests. I'm very supportive of this idea.

test/wrb.js
guidiaz: Given a queryId, it doesn't make too much sense to check whether its request contract changed its bytecode after a response was already reported. 

The only place in which this check makes sense would be when calling `WRB.readRequestBytecode(queryId)` before a query gets solved (as if it was already solved, it should revert with "not in posted status"). So:
- this test will be moved to a more appropiate place
- implementation of `readRequestBytecode` needs to be revisited

In other words, I believe it should not matter at all if a request contract modifies its bytecode at any time after being posted to the WRB, as it will be certain that any eventual response will correspond to the bytecode the request had at the moment the query got posted. This consistency check was done in a previous version of `reportResult`, but not any more.

contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
guidiaz: Great. However, some more changes to the data model will be proposed as to prevent this PR from becoming a breaking change (and therefore, forcing to either deploy new proxy address, or making 100% sure there are no pending requests in the WRB before upgrading it)

aesedepece: That sounds like music to my ears! :heart_eyes:

guidiaz: Done in https://github.com/witnet/witnet-solidity-bridge/pull/216/commits/f228c5953a91084d73c8e580ccae815f7741f9ad.

aesedepece (approved):

aesedepece (approved):

PR #217, opened 2021-12-24
Feat: Reduce pricefeed ids to 4 bytes

guidiaz:
- Shorter to code
- Simpler to remember
- Big enough to avoid collisions within same WitnetPriceRouter

```solidity
import "witnet-solidity-bridge/contracts/interfaces/IWitnetPriceRouter.sol`

contract MyContract {

  function lastBtcUsdValue() external view returns (int, uint, uint) {
     return IWitnetPriceRouter(<0x-address of WitnetPriceRouter>).valueFor(bytes32(0x24beead4));
  }

  function lastEthUsdValue() external view returns (int, uint, uint) {
     return IWitnetPriceRouter(<0x-address of WitnetPriceRouter>).valueFor(bytes32(0x3d15f701));
  }
}
```

### DIFF
--- a/contracts/examples/WitnetPriceRouter.sol
+++ b/contracts/examples/WitnetPriceRouter.sol
@@ -22,7 +22,7 @@ contract WitnetPriceRouter
         string  quote;
     }    
 
-    mapping (bytes32 => Pair) internal __pairs;
+    mapping (bytes4 => Pair) internal __pairs;
     mapping (address => bytes32) internal __pricefeedId_;
 
     bytes32[] internal __supportedCurrencyPairs;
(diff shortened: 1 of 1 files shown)


aesedepece (commented):
This is interesting! As far as I understand, this will help us save some gas (plus storage collateral on Conflux) while not breaking ERC2362 functionality. Right?

guidiaz:
This proposal just pretend to shorten the identifiers developers/apps can use to address the pricefeeds. First 4 bytes of the id will just be enough.

For instance, doing:`
```solidity
IWitnetPriceRouter(<0x-address of WitnetPriceRouter>).valueFor(bytes32(0x24beead4));
```

Instead of:
```solidity
IWitnetPriceRouter(<0x-address of WitnetPriceRouter>).valueFor(0x24beead43216e490aa240ef0d32e18c57beea168f06eabb94f5193868d500946);
```

Both ways will return same values.

aesedepece (approved):

PR #222, opened 2022-03-02
Feat(WRB): Allow batch reporting

guidiaz:
From authorized bridge nodes.

### DIFF
--- a/contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
+++ b/contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
@@ -188,6 +188,57 @@ abstract contract WitnetRequestBoardTrustableBase
     {
         _reportResult(_queryId, _timestamp, _drTxHash, _cborBytes);
     }
+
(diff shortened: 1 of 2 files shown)


contracts/impls/trustable/WitnetRequestBoardTrustableBase.sol
```diff
@@ -674,20 +755,22 @@ abstract contract WitnetRequestBoardTrustableBase
     // ================================================================================================================
     // --- Internal functions -----------------------------------------------------------------------------------------
 
-    function _reportResult(
+    function __reportResult(
```
Tommytrg, line 758: Why are we adding an extra `_`?

guidiaz: i'd rather prefer to prefix with two "__" any method that modifies storage. in a similar way, variables pointing to storage are also prefixed with dobule "__". this way it's much easier to follow source code and be aware when and how storage is either read or written, and therefore it eases the task of optimizing storage access, specially when dealing with loops or more complicated algorithms

PR #243, opened 2022-05-11
Fix: Witnet reward estimation on WRB Boba implementation

guidiaz:
See https://github.com/witnet/internal/issues/186#issuecomment-1123553140.


guidiaz:
Bridge Address balance after applying this patch:

image

Tommytrg (approved):

PR #278, opened 2022-09-09
refactor: remove already inherited interfaces

Tommytrg:
- `WitnetRequestBoardUpgradableBase` is already `Proxiable` because is inheriting `Upgradeable`and `Upgradeable` is `Proxiable`
- `WitnetRequestBoardTrustableBoba` is already `Payable` becasue is inherinting `WitnetRequestBoardTrustableBase` and `WitnetRequestBoardTrustableBase` is `Payable`


contracts/impls/WitnetRequestBoardUpgradableBase.sol
```diff
@@ -17,7 +17,6 @@ import "./WitnetProxy.sol";
 /// @author The Witnet Foundation.
 abstract contract WitnetRequestBoardUpgradableBase
     is
-        Proxiable,
```
aesedepece, line 20: Which of `Upgradable` or `WitnetRequestBoard` is `Proxiable`? :thinking:

PR #287, opened 2022-10-25
Chore: Adopt standard interfaces and patterns from Openzeppelin

### LINKED ISSUE #256: IWitnetPriceRouter should link to OpenZeppelin's IERC165 definition
- This is not a breaking-change
- New 0.5.9 tag should be created

guidiaz:
Close #256.


contracts/interfaces/IERC165.sol
```diff
@@ -1,24 +1,3 @@
...
-     * @dev Returns true if this contract implements the interface defined by
-     * `interfaceId`. See the corresponding
-     * https://eips.ethereum.org/EIPS/eip-165#how-interfaces-are-identified[EIP section]
-     * to learn more about how these ids are created.
-     *
-     * This function call must use less than 30 000 gas.
-     */
-    function supportsInterface(bytes4 interfaceId) external view returns (bool);
-}
+import "@openzeppelin/contracts/utils/introspection/IERC165.sol";
```
aesedepece, line 3: Interesting to see that we're re-exporting here instead of removing the interfaces altogether. I wouldn't care much about backwards compatibility as this is being shipped along a version bump.

guidiaz: Yeah, but tt's not only because backwards compatibility. The idea is to let the contents of contracts/interfaces/*.sol gather in one single place the whole interface set being either used or inherited by contracts within contracts/**.sol.

aesedepece (approved):

PR #293, opened 2022-11-08
Refactor: Enable current implementation coexist w/ V2 to come

guidiaz:


aesedepece:
Can you provide a little more color on how these changes help to achieve that coexistence? Thanks!

guidiaz:
Sure thing:
- There'll be a new contracts/WitnetRequestBoardV2.sol 
- There'll be a new and libs/WitnetV2.sol having new structs, enums and helpers functions, and we want this one to reuse Witnet.lib, and we want former cbor and parsing libs to not be bound to libs/Witnet.sol
- We're converting all methods in libs/WitnetDecoderLib.sol to internal, so there'll be no need to deploy it, and no need either to link it to WitnetParserLib.sol before deploying this one
- Most probably, there'll be new interfaces for encoding dr parameters, as well as for decoding dr results, that may require eventual dependency to new libs/WitnetV2.sol to come; because using "WitnetDecoderLib" and "WitnerParserLib" names for former libs would turn out rather misleading (they all were, i believe), new names proposed would be "WitnetCBOR" and "WitnetLib" the new names
- Actual implementations of WitnetRequestBoard (...Trustable...) and WitnetRequestBoardV2 (...Trustless...) will inherit from common contracts/impls/WitnetUpgradableBase.sol, which is new and replaces former contracts/WitnetRequestBoardUpgradableBase.sol"
- Witnet Solidity Bridge v2 will require not only implementation of WRBv2 and ...Trustless... contracts, but also abstract/concrete implementations for the Block Relay contract (e.g. WitnetBlocks) and, eventually, the Data Request Registry thing (e.g. WitnetBytecodes); this is why subfolders under contracts/impls have been moved under contracts/impls/boards

aesedepece:
>     * We're converting all methods in libs/WitnetDecoderLib.sol to internal, so there'll be no need to deploy it, and no need either to link it to WitnetParserLib.sol before deploying this one.
That definitely makes things easier!

guidiaz:
Supeseded by #317.

PR #363, opened 2023-08-03
Feat/0.7.10

guidiaz:
- Support methods in IWitnetRequestBoardDeprecating being used by WitnetRandomness prior to v0.6.1


contracts/libs/Witnet.sol
```diff
@@ -355,7 +355,7 @@ library Witnet {
         ) {
             return sha256(_bytecode);
         } else {
-            return 0x0;
+            return keccak256(_bytecode);
```
aesedepece, line 358: I think we can leverage the _De Morgan's laws_ to turn this into a positive expression:

```solidity
if (block.chainid == 1101 || block.chainid == 1442) {
    return keccak256(_bytecode);
} else {
    return sha256(_bytecode);
}
```

This is in fact not only more readable but also cheaper in gas terms, because the first comparison will short-circuit the `if` statement earlier than in the other approach.

aesedepece (changes requested):

contracts/libs/Witnet.sol
guidiaz: Moreover, we should work on a new WitnetRandomness implementation that relies on latest WitnetRequest implementation instead of WitnetMalleableBase, so we can deprecate (and remove from repository) the WitnetMalleableBase itself.

guidiaz: Just created a new issue for this: 
- https://github.com/witnet/witnet-solidity-bridge/issues/364

aesedepece: Thanks. All of that makes a lot of sense.

aesedepece (approved):

PR #407, opened 2024-02-01
Feat/modularize

guidiaz:


contracts/core/WitnetProxy.sol
```diff
@@ -84,24 +84,34 @@ contract WitnetProxy {
...
-        (bool _wasInitialized,) = _newImplementation.delegatecall(
+        (bool _wasInitialized, bytes memory _returnData) = _newImplementation.delegatecall(
             abi.encodeWithSignature(
                 "initialize(bytes)",
                 _initData
             )
         );
-        require(_wasInitialized, "WitnetProxy: unable to initialize");
+        if (!_wasInitialized) {
+            if (_returnData.length < 68) {
```
Tommytrg, line 94: Why 68? Could we add a comment here?

src/utils.js
```diff
@@ -0,0 +1,144 @@
...
+  traceTx,
+}
+
+function fromAscii (str) {
+  const arr1 = []
+  for (let n = 0, l = str.length; n < l; n++) {
+    const hex = Number(str.charCodeAt(n)).toString(16)
+    arr1.push(hex)
+  }
+  return "0x" + arr1.join("")
```
Tommytrg, line 28: We can achieve the same behaviour with `Array.from`
```suggestion
  return "0x" + Array.from(str, char => char.charCodeAt(0).toString(16)).join("");
```

contracts/core/defaults/WitnetRequestBytecodesDefault.sol
```diff
@@ -370,35 +288,40 @@ contract WitnetBytecodesDefault
...
+    // function lookupRadonSLA(bytes32 _slaHash)
+    //     external view
+    //     override
+    //     returns (Witnet.RadonSLA memory sla)
+    // {
+    //     sla = __database().slas[_slaHash];
+    //     if (sla.numWitnesses == 0) {
+    //         revert UnknownRadonSLA(_slaHash);
+    //     }
+    // }
```
Tommytrg, line 321: Can we remove this comment?
```suggestion
```

contracts/interfaces/IWitnetRequestBytecodes.sol
```diff
@@ -0,0 +1,65 @@
...
+            uint16 resultMaxSize,
+            string[][] calldata args
+        ) external returns (bytes32 radHash);
+
+    function totalDataProviders() external view returns (uint);
+
+    /// ===============================================================================================================
+    /// --- Deprecating methods ---------------------------------------------------------------------------------------
+
+    // function lookupRadonSLA(bytes32 slaHash) external view returns (Witnet.RadonSLA memory);
```
Tommytrg, line 63: What's the benefit of keeping this comment?

contracts/patterns/Payable.sol
```diff
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: MIT
-
 pragma solidity >=0.6.0 <0.9.0;
 
-import "../interfaces/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
Tommytrg, line 4: Could we also remove the extra file for the rest of openzeppelin contracts?
- ERC165
- Initializable
- Ownable
- ReentrancyGuard

package.json
```diff
@@ -1,68 +1,86 @@
...
-    "witnet",
-    "ethereum"
-  ],
-  "license": "MIT",
   "dependencies": {
-    "@openzeppelin/contracts": "4.8.0",
-    "@openzeppelin/contracts-upgradeable": "4.8.0",
-    "ado-contracts": "1.0.0",
-    "lodash": "4.17.21"
+    "dotenv": "^16.4.1",
```
Tommytrg, line 57:
```suggestion
    "dotenv": "^16.4.4",
```

src/utils.js
```diff
@@ -0,0 +1,144 @@
...
+      })
+    rl.on("close", function () {
+      resolve()
+    })
+  })
+  return answer
+}
+
+async function readJsonFromFile (filename) {
+  lockfile.lockSync(filename)
```
Tommytrg, line 116: Why do we need `lockfile`?

package.json
```diff
@@ -1,68 +1,86 @@
...
-    "truffle": "~5.11.5",
-    "truffle-assertions": "0.9.2",
-    "truffle-flattener": "~1.6.0",
-    "truffle-plugin-verify": "~0.6.2"
+    "eslint": "^8.56.0",
+    "eslint-config-standard": "^17.1.0",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-n": "^16.6.2",
+    "eslint-plugin-promise": "^6.1.1",
+    "eth-gas-reporter": "^0.2.27",
```
Tommytrg, line 74: How are we using` eth-gas-reporter`

package.json
```diff
@@ -1,68 +1,86 @@
...
-    "solidity-coverage": "0.7.16",
-    "truffle": "~5.11.5",
-    "truffle-assertions": "0.9.2",
-    "truffle-flattener": "~1.6.0",
-    "truffle-plugin-verify": "~0.6.2"
+    "eslint": "^8.56.0",
+    "eslint-config-standard": "^17.1.0",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-n": "^16.6.2",
+    "eslint-plugin-promise": "^6.1.1",
```
Tommytrg, line 73: I think we aren't using this `eslint` files but `eslint-config-standard`, could we remove the other ones?

package.json
```diff
@@ -1,68 +1,86 @@
...
+    "eth-gas-reporter": "^0.2.27",
+    "eth-helpers": "^1.3.0",
+    "hardhat": "^2.19.5",
+    "js-sha256": "^0.11.0",
```
Tommytrg, line 77: Are we using this dependency?
```suggestion
```

package.json
```diff
@@ -1,68 +1,86 @@
...
+    "nanoassert": "^2.0.0",
+    "sha3-wasm": "^1.0.0",
+    "solhint": "^4.1.1",
+    "truffle": "^5.11.5",
+    "truffle-assertions": "^0.9.2",
```
Tommytrg, line 82: Are we using this dependency?
```suggestion
```

package.json
```diff
@@ -1,68 +1,86 @@
...
+    "truffle-flattener": "^1.6.0",
+    "typescript": "latest"
```
Tommytrg, line 84: Do we have any typescript file?
```suggestion
```

contracts/libs/Create3.sol
```diff
@@ -0,0 +1,113 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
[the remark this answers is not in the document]
Tommytrg: The ones that include lib in the name, can be deployed and and have public methods

contracts/patterns/Payable.sol
Tommytrg: Or if we want to keep a file for each of the contracts from openzeppelin, do the same for IERC20.sol

package.json
Tommytrg: Added use in latest commit

contracts/core/defaults/WitnetRequestFactoryDefault.sol
```diff
@@ -20,15 +17,16 @@ contract WitnetRequestFactoryDefault
         WitnetRequestFactoryData,
         WitnetUpgradableBase        
 {
-    using ERC165Checker for address;
+    /// @notice Reference to Witnet Data Requests Bytecode Registry.
+    WitnetRequestBytecodes immutable public override(WitnetRequestFactory, WitnetRequestTemplate) registry;
 
-    /// @notice Reference to Witnet Data Requests Bytecode Registry
-    WitnetBytecodes immutable public override(IWitnetRequestFactory, WitnetRequestTemplate) registry;
+     /// @notice Reference to the Witnet Request Board that all templates built out from this factory will refer to.
```
Tommytrg, line 23:
```suggestion
    /// @notice Reference to the Witnet Request Board that all templates built out from this factory will refer to.
```

Tommytrg (approved):

PR #411, opened 2024-02-21
Chore/2.0.3

guidiaz:


contracts/patterns/Upgradeable.sol
```diff
@@ -4,6 +4,8 @@
 
 pragma solidity >=0.6.0 <0.9.0;
 
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
```
Tommytrg, line 7: Why do we need to import `ERC165`?

guidiaz: Weird, but yes, otherwise hardhat compile upon package prepare fails

PR #443, opened 2024-10-02
Chore/lighten wit oracle base

guidiaz:


contracts/core/trustless/WitOracleRadonRegistryDefault.sol
```diff
@@ -315,39 +318,82 @@ contract WitOracleRadonRegistryDefault
...
 
     function verifyRadonRetrieval(
             Witnet.RadonRetrievalMethods _requestMethod,
-            string calldata _requestURL,
-            string calldata _requestBody,
+            string memory _requestURL,
+            string memory _requestBody,
             string[2][] memory  _requestHeaders,
-            bytes calldata _requestRadonScript
+            bytes memory _requestRadonScript
```
Tommytrg, line 396: Is it necessary to replace `calldata` with `memory`? Isn't `calldata` cheaper than `memory`?

contracts/core/trustless/WitOracleRadonRegistryDefault.sol
```diff
@@ -486,17 +587,20 @@ contract WitOracleRadonRegistryDefault
...
             
             // Build radon retrieval bytecode:
             bytes memory _bytecode = _retrievals.encode(
-                _retrieveArgs,
-                _aggregator.encode(),
-                _tally.encode(),
-                0 //_resultMaxSize
+                _retrieveArgsValues, 
+                _aggregateReducer.encode(), 
+                _tallyReducer.encode(),
```
Tommytrg, line 602:
```suggestion
                _aggregateReducer.encode(),
                _tallyReducer.encode(),
```

contracts/interfaces/IWitOracleRequestFactory.sol
```diff
@@ -39,7 +62,28 @@ interface IWitOracleRequestFactory {
...
             Witnet.RadonReducer calldata aggregate,
             Witnet.RadonReducer calldata tally
-        ) external returns (address template);
+        ) external returns (IWitOracleRequestTemplate);
+
+    /// @notice Builds a modal WitOracleRequestTemplate instance out of one single parameterized 
+    /// @notice Witnet Radon Retrieval. Modal request templates produce Witnet-compliant data requests
+    /// @notice that apply the same data retrieval script to multiple data providers that are expected to
+    /// @notice produce exactly the same result. Moreover, data requests built out of a modal request templates
+    /// @notice apply a Mode reducing function at both the Aggregate and Tally stages everytime they get
```
Tommytrg, line 71:
```suggestion
    /// @notice apply a Mode reducing function at both the Aggregate and Tally stages every time they get
```

contracts/interfaces/IWitOracleRequestFactory.sol
```diff
@@ -39,7 +62,28 @@ interface IWitOracleRequestFactory {
...
+    /// @notice eventually solved on the Wit/Oracle blockchain. You can optionally specify a list of filters 
+    /// @notice to be applied at the Tally resolution stage (i.e. witnessing nodes on the Wit/Oracle blockchain
+    /// @notice reporting results that get ultimately filtred out on the Tally stage would get slashed by losing collateral).
```
Tommytrg, line 74:
```suggestion
    /// @notice reporting results that get ultimately filtered out on the Tally stage would get slashed by losing collateral).
```

contracts/core/trustless/WitOracleRadonRegistryDefault.sol
guidiaz: > Is it necessary to replace `calldata` with `memory` ?
Yep, as `verifyRadonRetrieval(uint8,string,string,string[][],bytes)` is now internally called from the implementation of the (new) `verifyRadonRetrieval(bytes32,string)` which is getting _requestURL, _requestBody, _requestHeaders and `_requestRadonScript` as modified memory buffers as returned from `lookupRadonRetrieval(bytes32)`.

> `Isn't `calldata` cheaper than `memory` ?
It depends. Not necessarily. In some cases using `calldata` parameters may slightly reduce the gas consumption, but not in this case whatsoever, as formerly `calldata` parameters had to be copied to storage anyway.

Tommytrg (approved):

PR #456, opened 2025-01-15
Fix/determine supported networks

guidiaz:


settings/artifacts.js
```diff
@@ -17,11 +17,6 @@ module.exports = {
       WitPriceFeedsDataLib: "WitPriceFeedsDataLib",
     },
   },
-  "polygon:amoy": {
```
Tommytrg, line 20: What has changed in polygon amoy?

guidiaz: nothing really, possibly just upgraded to latest compilation

Tommytrg (approved):

PR #471, opened 2025-07-22
Chore/optimize bech32

guidiaz:


contracts/libs/Witnet.sol
```diff
@@ -1099,19 +1099,13 @@ library Witnet {
 
     {
         bytes memory _bytes = new bytes(64);
-        for (uint8 _i = 0; _i < _bytes.length;) {
-            uint8 _f = uint8(_bytes32[_i/2] & 0x0f);
-            uint8 _l = uint8(_bytes32[_i/2] >> 4);
-            _bytes[_i ++] = _toHexChar(_f);
-            _bytes[_i ++] = _toHexChar(_l);
+        for (uint8 _i; _i < _bytes.length;) {
```
Tommytrg, line 1102: Is this a refactor? As we are using _i++, the resulting bytes are not equivalent

Tommytrg (approved):

